PR #3, opened 2018-08-01
Update __init__.py

znicholls:
Why do we need get_version twice? Can we delete one or add a comment explaining why?


rgieseke:
Copy and paste error ... that's probably why one should have code review :-)

Issue #9, opened 2018-08-03
Makefile


znicholls:
It would be good to make this now and stick to a few conventions. I often find that it's good to have

`make test` - install all dependencies etc., activate `venv`/use `venv` packages (if not already) and run all the tests

```
test: venv
    ./venv/bin/pytest tests
```

`make venv` - install a virtualenv for development (i.e. install `openscm` with `pip install -e`) (you know this one)

`make clean` - in this case clean our `venv`?

`make test-upload-pypi`, `make-upload-pypi` - the way these are done in `pymagicc` is wicked @rgieseke, can you repeat the dose here?

`make notebook` (could be handy)

```
notebook: venv
    ./venv/bin/jupyter notebook
```

or maybe 

```
notebook: venv
    source venv/bin/activate; jupyter notebook
```

PR #31, opened 2018-10-29
Add makefile

rgieseke:
I'm leaving out the notebook task we have in mentioned #9 as we don't have notebooks yet.


Makefile
```diff
@@ -0,0 +1,46 @@
...
+	$(eval TEMPVENV := $(shell mktemp -d))
+	python3 -m venv $(TEMPVENV)
+	$(TEMPVENV)/bin/pip install pip --upgrade
+	$(TEMPVENV)/bin/pip install openscm
+	$(TEMPVENV)/bin/python -c "import sys; sys.path.remove(''); import openscm; print(openscm.__version__)"
+
+clean:
+	rm -rf venv
+
+.PHONY: clean test black clean flake8 docs
```
swillner: shouldn't `publish-on-pypi` and `test-pypi-install` also be `PHONY`?

```suggestion
.PHONY: clean test black clean flake8 docs publish-on-pypi test-pypi-install
```

setup.cfg
```diff
@@ -5,3 +5,7 @@ versionfile_source = openscm/_version.py
 versionfile_build = openscm/_version.py
 tag_prefix = v
 parentdir_prefix = openscm-
+
+[flake8]
+max-line-length = 88
+ignore = E203, E266, E501, W503
```
znicholls: can we add the following explanation

```
# E203 - whitespace before ':'. Opposite convention enforced by black so ignore here
# E266 - too many leading '#' for block comments, dumb rule so ignore
# E501 - line too long. Handled by black, we have longer lines
# W503 - line break before binary operator. Dumb rule, ignore
```

rgieseke: Pushed to other branch, whoopsie ...

PR #35, opened 2018-10-29
Add units

### LINKED ISSUE #28: List of common units
@znicholls has a list, will put somewhere ...

znicholls:
`notebooks/emissions-units-with-pint.ipynb` gives an example of how this works with Pint


notebooks/emissions_units.txt
```diff
@@ -0,0 +1,58 @@
...
+N2O = 14/44 * N = nitrous_oxide_emissions
+NOx = 14/46 * N = nox_emissions = NOX
+N2ON = 14/28 * N = nitrous_oxide_farming_style_emissions
+NH3 = 14/17 * N = ammonia_emissions
+
+S = [sulfur_emissions]
+SO2 = 32/64 * S = sulfur_dioxide_emissions
+SOx = SO2 = SOX
+
+VOC = [VOC_emissions] = NMVOC = non_methane_volatile_organic_compouns_emissions
```
rgieseke:
```suggestion
VOC = [VOC_emissions] = NMVOC = non_methane_volatile_organic_compounds_emissions 
```

rgieseke:
Haha, before i can even comment the CEDS stuff is gone :-)

notebooks/emissions_units.txt
znicholls: can you suggest again? Good suggestion just in the wrong file now

openscm/definitions/emissions_units.txt
```diff
@@ -0,0 +1,58 @@
...
+N2O = 14/44 * N = nitrous_oxide_emissions
+NOx = 14/46 * N = nox_emissions = NOX
+N2ON = 14/28 * N = nitrous_oxide_farming_style_emissions
+NH3 = 14/17 * N = ammonia_emissions
+
+S = [sulfur_emissions]
+SO2 = 32/64 * S = sulfur_dioxide_emissions
+SOx = SO2 = SOX
+
+VOC = [VOC_emissions] = NMVOC = non_methane_volatile_organic_compouns_emissions
```
rgieseke:
```suggestion
VOC = [VOC_emissions] = NMVOC = non_methane_volatile_organic_compounds_emissions 
```

rgieseke:
To discuss:
Hyphens or Underscores in unit names:

Maybe "N2O_N"

N2O-N is not possible I think, they need to be valid Python variable names.

znicholls:
My vote would be neither. Removed ambiguity about whether an underscore is
really an underscore or whether it's meant to be a hyphen. Option b is to
have a convenience function for conversion which understands hyphens,
strips them, gets the conversions and then returns the results with hyphens
back in.

dev-requirements.txt
```diff
@@ -3,3 +3,10 @@ twine>=1.11.0
 wheel>=0.31.0
 black
 flake8
+notebook
+git+https://github.com/hgrecco/pint.git
```
swillner: pint is not available via pypi?

openscm/definitions/__init__.py
```diff
@@ -0,0 +1,9 @@
+from os.path import join, dirname
+
+
+import pint
+
+unit_registry = pint.UnitRegistry()  # start a unit repository using the default variables
+# should move the definitions somewhere more stable
+# openscm/definitions ?
+unit_registry.load_definitions(join(dirname(__file__), 'emissions_units.txt'))
```
swillner: Can we avoid parsing a file and define the units in-code?

openscm/definitions/emissions_units.txt
```diff
@@ -0,0 +1,58 @@
+yr = 1 * year
```
swillner: I suppose we should also add "a" (annum) as that seems to be unknown to pint...
```suggestion
a = annum = yr = 1 * year
```

openscm/definitions/emissions_units.txt
```diff
@@ -0,0 +1,58 @@
...
+kt = 1000 * t
+
+C = [carbon_emissions]
```
swillner: Shouldn't we define the actual weight as units, e.g. "tC" and "gC"? That way we can avoid the otherwise necessary space between "t" and "C" and are actually defining units?
I would strongly suggest that.

swillner:
In general, are the units we would actually use so complex we need a sophisticated library for that?

swillner:
Hmmm, having looked into it, I am not that happy with using Pint. If I have two units and want to do lots of conversions between the two, I cannot get just a (pre-/post-) offset and a scaling factor (without hacking Pint or resorting to calculating these myself from some chosen points). Anything else would be inefficient, at least in the low-level API... What can we do about that?
Where do you need conversion in the high-level API?
What units do we need actually?

swillner:
Alternatively, we could have a look at unyt, which seems to have the functionality I want: https://unyt.readthedocs.io/en/latest/modules/unyt.unit_object.html#unyt.unit_object.Unit.get_conversion_factor

dev-requirements.txt
rgieseke: I think this was just to be able to install latest version with Pandas support

openscm/definitions/__init__.py
rgieseke: It's possible to do these definitions programmatically, but not the library itself uses textfiles.

rgieseke:
@swillner What unit conversion is happening in the low-level API?

openscm/definitions/emissions_units.txt
znicholls: I thought about this too. However I think that blends two things which are actually separate (whatever the outcome, I agree that this conversation needs to go in the docs somewhere).

For example, when you buy bananas, you buy a kilo of bananas. The units are kg bananas, not kgbananas. I think an emissions unit is three separate components, the mass, what the stuff is and the time period (/yr, /day etc.). 

Blending the first two means that we have to define each of the masses for every new gas e.g. "tC, MtC, GtC, PgC...", "tCO2, MtCO2, GtCO2, PgCO2" and the conversions between all of these things. This will be a real pain and hard to keep track of as the number of combinations is much bigger because you have to two things to keep track of (mass and the stuff) rather than keeping them separate and letting Pint do the combinations for us.

I also think blending the two makes it less clear what these emissions units actually are. When you think about it, they're super weird. They're not simply atomic masses, if they were one could convert between C and N2O with no issue. They're really a counting measure which is particular to emissions. Hence we have this funny (or at least it seems funny to me now) convention where C and CO2 are interchangeable (despite being fundamentally different species), but C and N2O are not. In any case, if we keep the mass separate from the emissions, it's easy for us to say that you can swap C and CO2 but not e.g. C and N2O (and hence CO2 and N2O by consequence thanks to Pint's handling). It then also makes it much easier to work with contexts as we only need to define gas to gas conversions and don't need to define the conversions between all the masses of all the different gases for each new context.

znicholls:
> In general, are the units we would actually use so complex we need a sophisticated library for that?

I would argue yes. People (me) have cocked them up enough to mean that if we can just do the unit handling once, in a fairly clever way, we should.

znicholls:
> Hmmm, having looked into it, I am not that happy with using Pint.

It's the only unit library with Pandas support. For me that's reason enough to use it.

> What can we do about that?

I'm happy to make a PR to Pint this week which gives you a convenience function to Pint which does this (or put it first in OpenSCM, then make a PR in Pint)

```python
>>> import pint
>>> ureg = pint.UnitRegistry()
>>> a = 1*ureg.m
>>> a
<Quantity(1, 'meter')>
>>> a.to("km").magnitude
0.001
```

It would look like this

```python
def get_conversion_factor(start_units, other_units):
    # type checking etc goes here
    pint_quantity = 1 * ureg(start_units)
    return pint_quantity.to(other_units).magnitude
```

We could then cache those values or whatever you wanted.

znicholls:
> Where do you need conversion in the high-level API?

If people want to make plots in different units, that should be easy (I can't think of any other use cases...).

> What unit conversion is happening in the low-level API?

All of the conversion when we get requests from models e.g. MAGICC's adaptor says, give me CO2 emissions in GtC/yr but the user set them as MtC / month, the low-level API has to do that conversion.

znicholls:
> What units do we need actually?

Ideally all the emissions units in that file, plus conversions for all the key IPCC metrics (AR5GWP20, AR5GWP100, AR5GTP20, AR5GTP100, AR4GWP100, AR4GWP20 etc.), concentrations units (ppm, ppb, ppt), physical units (K, C, W/m^2, kg, m) etc.

swillner:
Hmm, ok, we shouldn't use two different unit libraries... Then I will use such a function getting offset and scaling from points, though far from optimal. But including something like that in Pint would probably better rely on Pint internals...

openscm/definitions/emissions_units.txt
swillner: The prefixes (M, G, P, ...) should be handled be the unit library (Pint also does so, I hope?!), so I don't see a problem there. Can we otherwise circumvent forcing people to put a space between weight unit and gas "unit"?

openscm/definitions/__init__.py
swillner: Hmm, I hate parsing text files, but ok ;)

openscm/definitions/emissions_units.txt
swillner: I always thought "a" is much more common than "yr"...
Also, we should add "h" as for "hour" (see https://github.com/hgrecco/pint/issues/719)

openscm/definitions/emissions_units.txt
znicholls: > Can we otherwise circumvent forcing people to put a space between weight unit and gas "unit"?

Sure. We just wrap whatever unit conversions we want to do with something like 

```python
def convert_unit(in_unit, out_unit):
    if in_unit.startswith("t", "Mt", "Gt", "g", "Pg", "Mg", "Gg"):  # etc.
        in_unit = "{} {}".format(in_unit[:2], in_unit[2:])
    # the rest
```

Having said that, I still think this sends the wrong signal. They're not one unit, they're two.

znicholls:
> though far from optimal

I don't understand why? If we implement something in Pint which is exactly the same as the unyt functionality, isn't that fine?

swillner:
unyt directly exposes the offset and scaling that are used internally (in Pint these are hidden much deeper). Using two points (e.g. 1 unit and 10 unit) to calculate offset and scaling is just a work-around. Maybe it's just my desire for asthetic code then ;)
At best, I would want a unit library to expose something like a "Converter" class, that takes two units and exposes functions to convert numbers (just floats, not some wrapping Unit classes) back and forth between the two units...

openscm/definitions/emissions_units.txt
swillner: Hmm, you're right, then let's follow the philosophy of the gas type as a separate unit... (`convert_unit` would only take care of some weights and the very beginning of the full unit though...)

openscm/definitions/emissions_units.txt
znicholls: can you suggest that?

openscm/definitions/emissions_units.txt
znicholls: yep `convert_unit` would need some serious testing for us to be sure it covers all the uses cases, I'm happy for that to go in the bin

swillner:
Ok, I will have a look if we can make a PR in Pint with something like that, what do you think?

znicholls:
> Ok, I will have a look if we can make a PR in Pint with something like that, what do you think?

I like it. Good procrastination ;)

znicholls:
@rgieseke to you for final say and then merge?

Is this all we needed in this PR?

openscm/definitions/emissions_units.txt
swillner: Idea for a compromise: We could define `gC = g * C`, `tC = t * C` et al (so two additional lines per new gas unit)...

znicholls: genius! We could also define those units dynamically... Let me have a fiddle today/tomorrow

rgieseke: Awesome, that looks great! And feel more natural and like what people are already using.

openscm/definitions/openscm_units.txt
```diff
@@ -0,0 +1,64 @@
+# syntax is either:
+# for variables which relate to existing variables in the registry
+# name = relation to unit already in registry = other names
+# for variables which should define a new base unit
+# name = [long_name] = other names
+
+a = 1 * year = annum = yr
+kt = 1000 * t
```
swillner:
```suggestion
h = hour
# since kt is used for "knot" in the defaults:
kt = 1000 * t
```

openscm/definitions/emissions_units.txt
swillner: done ;)

openscm/definitions/emissions_units.txt
znicholls: Other question here then, do we want to still use the default registry or should we restrict only to units we define to avoid unexpected collisions (like the weird thing with h being planck constant by default?)

swillner: Hmm, is there an SI registry? It's nice to have all the standard units in there already... Otherwise, let's just overwrite the default where needed.
We should have a list of units we want to be supported by the OpenSCM standard (e.g., for other implementations) in the docs/the standard.

openscm/definitions/openscm_units.txt
rgieseke: Do we need the defaults like "knot" at all?

openscm/definitions/emissions_units.txt
rgieseke: +1 for sensible defaults and no weird units (if that means we only use ours by default that's fine I'd say)

openscm/definitions/openscm_units.txt
```diff
@@ -0,0 +1,66 @@
...
+# for variables which relate to existing variables in the registry
+# name = relation to unit already in registry = other names
+# for variables which should define a new base unit
+# name = [long_name] = other names
+
+a = 1 * year = annum = yr
+h = hour
+# since kt is used for "knot" in the defaults:
+kt = 1000 * t
+
```
swillner: I just found that `°C` and `°F` are note defined (but `°` is converted to `degree`):

```suggestion
degreeC = degC
degreeF = degF
```
although @rgieseke opposes ridiculous units such as `°F` ;) rightly!

openscm/definitions/emissions_units.txt
znicholls: > Hmm, is there an SI registry?

I don't think so, we could split it out in Pint (again not hard, just more fiddling)

> We should have a list of units we want to be supported by the OpenSCM standard (e.g., for other implementations) in the docs/the standard.

Good idea

> Otherwise, let's just overwrite the default where needed.

Cool

rgieseke:
@swillner @znicholls In general, is this ready to be merged or are you still tweaking?

rgieseke:
Current failure is from the notebook.

openscm/definitions/emissions_units.txt
swillner: @znicholls fyi, I have some time right now and will quickly include the additional units...

znicholls:
> In general, is this ready to be merged or are you still tweaking?

Nope. Let's let Sven do his additions then I'll do the usual:

- add tests
- add docs
- add changelog entry
- add example

swillner:
I'm not yet done with the units... Since it's just adding two lines (g and t) for each gas, should'nt we resort to defining them programmaitcally, e.g.
```
gases = {
    "C": "carbon,
    ...
}
for k, v in gases.items():
    ureg.define("{} = [{}]".format(k, v))
    ureg.define("g{} = g * {}".format(k, k))
    ureg.define("t{} = t * {}".format(k, k))
```
and then add the conversions between them?

znicholls:
> should'nt we resort to defining them programmaitcally

yes I think you've now won the programatically defining 'battle'

swillner:
and should `C = [carbon]` or `C = [carbon_emissions]`? I would prefer the first and then `gC = [carbon_emissions]`...

znicholls:
> and should `C = [carbon]` or `C = [carbon_emissions]`? I would prefer the first and then `gC = [carbon_emissions]`

Good question

- `t C / yr` is really carbon emissions
    - emissions are, after all, a flux and `t C` is cumulative emissions
- hence `CO2 = [carbon_dioxide]` and then `C = 12 / 44 * CO2`
    - i.e. no new name
        - for our context C is always carbon dioxide 
        - this means you can't give e.g. methane in units of C but I think that's ok and the only way to avoid confusion
- `gC = g * C`
    - again no name, if we have to have a name we could use `[gram_carbon_dioxide]` but I think that's overkill

setup.py
```diff
@@ -37,13 +37,13 @@
     keywords="simple climate model",
     license="GNU Affero General Public License v3.0 or later",
     packages=["openscm"],
-    tests_require=["pytest"],
+    install_requires=["pyam-iamc"],
```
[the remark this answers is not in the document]
swillner: shoulg'nt that be `pint`?

swillner:
Are there more definitions coming? Shouldn't we move these to a file `units.py` (either directly in `openscm` or `openscm.definitions`)? I will extend (or create) `units.py` with a `UnitConverter` class.

znicholls:
> Are there more definitions coming?

Possibly stuff like what all our regions actually are, what the variables are etc. etc. In any case I think splitting to having ``openscm.units`` or ``openscm.definitions.units`` is a great idea!

setup.py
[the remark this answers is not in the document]
znicholls: We need `pyam-iamc` for our high level dataframe.

It will be Pint, once the new release is done (or given we're not using units in dataframes yet we might be able to do it now.., I'll check in another PR).

swillner:
All right, that's it from my side...
@znicholls would you want to fix the notebooks (cf. travis)?

swillner:
@chrisroadmap @znicholls What about the https://github.com/openclimatedata/openscm/tree/unit_fiddle branch?

znicholls:
> What about the https://github.com/openclimatedata/openscm/tree/unit_fiddle branch?

Having discussed with @chrisroadmap, dead

znicholls:
> @znicholls would you want to fix the notebooks (cf. travis)?

yep will do this arvo

swillner:
awesome!

znicholls:
Done from my side :)

openscm/units.py
```diff
@@ -32,14 +68,15 @@
...
-# make private
-gases = {
+_gases = {
     "C": "carbon",
     "CO2": ["12/44 * C", "carbon_dioxide"],
     "N": "nitrogen",
     "N2O": ["14/44 * N", "nitrous_oxide"],
-    "NOx": ["14/46 * N", "nox"],
     "N2ON": ["14/28 * N", "nitrous_oxide_farming_style"],
+    # "NOx": ["14/46 * N", "nox"],
```
swillner: `Nox` won't be convertible to `N`?

swillner: ah, seeing in below in the tests ;)

tests/unit/test_units.py
```diff
@@ -1,11 +1,87 @@
...
+        unit_registry("carbon")
+
+
+def test_nitrogen():
+    N = unit_registry("N")
+    assert N.to("N2ON").magnitude == 28 / 14
+
+
+def test_nox():
+    # do we want to allow this?
```
swillner: @znicholls what do you mean?

tests/unit/test_units.py
```diff
@@ -1,11 +1,87 @@
...
+    assert CO2.to("C").magnitude == 12.0 / 44.0
+
+
+def test_alias():
+    CO2 = unit_registry("carbon_dioxide")
+    assert CO2.to("C").magnitude == 12.0 / 44.0
+
+
+def test_base_unit():
+    # I don't know why this is this way, anyway
```
swillner: `carbon` should be defined, right? checking...

openscm/units.py
```diff
@@ -0,0 +1,243 @@
...
+# Other definitions:
+
+unit_registry.define("a = 1 * year = annum = yr")
+unit_registry.define("h = hour")
+unit_registry.define("degreeC = degC")
+unit_registry.define("degreeF = degF")
+unit_registry.define("kt = 1000 * t")  # since kt is used for "knot" in the defaults
+
+unit_registry.define("ppt = [concentrations]")
+unit_registry.define("ppb = 1000*ppt")
```
swillner: just for style ;)
```suggestion
unit_registry.define("ppb = 1000 * ppt")
```

openscm/units.py
```diff
@@ -0,0 +1,243 @@
...
+unit_registry.define("ppm = 1000*ppb")
```
swillner:
```suggestion
unit_registry.define("ppm = 1000 * ppb")
```

swillner:
great work @znicholls !

swillner:
Closes #23

tests/unit/test_units.py
znicholls: If we let people convert NOx to N, then they can also convert NOx to N2O. Converting NOx to N2O doesn't make any sense and having that behaviour be allowed without at least a warning seems dangerous. Maybe it's something we can wrap in our converter class?

tests/unit/test_units.py
znicholls: ye I don't get it either..

znicholls:
> great work @znicholls !

Team effort (as they most likely all will be)!

tests/unit/test_units.py
swillner: Hmm, we could have warnings in the converter class, but that would be lot's of hacking, I am afraid (at least if we want a warning also for complex combinations of the units, e.g. with `kgN * m^2 / s` or so)

swillner: Either way, are those TODO comments or meant for the user?

tests/unit/test_units.py
swillner: Ahh, I know now: `C = [carbon]` tells Pint, that `C` is of *dimension* `carbon` (like `s` is of dimension `time`). So we could add `carbon` also as an alias for `C` (add `carbon = C`)?

znicholls: If that works, let's go for it?

tests/unit/test_units.py
znicholls: They're comments to prompt this discussion haha

Thinking about it more, I think the cleanest solution would be to have contexts for NOx <--> N and CH4 <--> C conversions. That way they're still possible, you just have to be sure you want to do them and hence are less likely to accidentally get things wrong.

tests/unit/test_units.py
swillner: done

tests/unit/test_units.py
swillner: cool ;)

tests/unit/test_units.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
```
znicholls: @swillner I generally try and put my imports in 3 levels, each separated by two blank lines:

- packages included with Python
- external packages
- modules from within the package

e.g.

```python
import os


import pytest
import numpy as np


from opensc.units import unit_registry
```

I don't actually mind but might be good to choose a convention and stick to it?

swillner: ah, ok. compromise: 1 line in between? ;)

znicholls: lol ok

Makefile
```diff
@@ -43,4 +46,4 @@ test-pypi-install: venv
 clean:
 	rm -rf venv
 
-.PHONY: clean test black clean flake8 docs publish-on-pypi test-pypi-install
+.PHONY: clean test test_all black clean flake8 docs publish-on-pypi test-pypi-install
```
rgieseke:
```suggestion
.PHONY: clean test test-all black clean flake8 docs publish-on-pypi test-pypi-install
```

rgieseke: Hyphenize for consistency.

dev-requirements.txt
```diff
@@ -1,5 +1,3 @@
-setuptools>=38.6.0
-twine>=1.11.0
-wheel>=0.31.0
-black
-flake8
+git+https://github.com/thedataincubator/expectexception.git@master
+# can't unpin this as then definition doesn't work properly...
```
rgieseke: It's not possible to install from git in `setup.py`? Otherwise we could get rid of dev-requirements?

openscm/units.py
```diff
@@ -0,0 +1,326 @@
+"""
+Unit handling.
+
+Our unit handling library is `Pint <https://github.com/hgrecco/pint>`_. This allows us
+to define units simply as well as providing us with the ability to define contexts.
+
+
+**A note on emissions units**
+
+Emissions are a flux composed of three parts: mass, the stuff being emitted and the
```
rgieseke:
```suggestion
Emissions are a flux composed of three parts: mass, the species being emitted and the
```

openscm/units.py
```diff
@@ -0,0 +1,326 @@
...
+However, it can be performed within the context 'CH4_conversions' as shown below:
+
+.. code:: python
+
+    >>> from openscm.units import unit_registry
+    >>> CH4 = unit_registry("CH4")
+    >>> CH4.to("C")
+    pint.errors.DimensionalityError: Cannot convert from 'CH4' ([methane]) to 'C' ([carbon])
+
+    # with a context, the conversion becomes legal again
```
rgieseke: Just curious, where are people actually expressing CH4 as C?

dev-requirements.txt
znicholls: unfortunately I don't think so (when I tried it blew up...)

openscm/units.py
znicholls: Good question, I can't remember where I've seen it (we can remove if you want but I think we've done the work now so just leave in?)

dev-requirements.txt
rgieseke: Alright, seems to be possible but difficult (`dependency_links`) -- so not worth fiddling around with, we could also put them directly into the Makefile to reduce the numbers of files and locations to check.

openscm/units.py
```diff
@@ -0,0 +1,326 @@
...
+- case-sensitivity. In order to provide a simplified interface, using all uppercase
+  versions of any unit is also valid e.g. ``unit_registry("HFC4310mee")`` is the same as ``unit_registry("HFC4310MEE")``
+- hyphens and underscores in units. In order to be Pint compatible and to simplify
+  things, we strip all hyphens and underscores from units.
+
+As a convenience, we allow users to combine the mass and the type of emissions to make
+a 'joint unit' e.g. "tCO2" but it should be recognised that this joint unit is a
+derived unit and not a base unit.
+
+By definining these three separate components, it is much easier to track what
```
rgieseke:
```suggestion
By defining these three separate components, it is much easier to track what
```

tests/unit/test_units.py
```diff
@@ -0,0 +1,120 @@
...
+import pytest
+
+from openscm.units import unit_registry
+from openscm.units import DimensionalityError, UndefinedUnitError, UnitConverter
+
+
+def test_unit_registry():
+    CO2 = unit_registry("CO2")
+    assert CO2.to("C").magnitude == 12.0 / 44.0
```
rgieseke:
```suggestion
    assert CO2.to("C").magnitude == 12 / 44
```

rgieseke: Use same accuracy as below?

Also, should this be some kind of almost_equal?

tests/unit/test_units.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def test_alias():
+    CO2 = unit_registry("carbon_dioxide")
+    assert CO2.to("C").magnitude == 12.0 / 44.0
```
rgieseke:
```suggestion
    assert CO2.to("C").magnitude == 12 / 44
```

tests/unit/test_units.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def test_ppt():
+    ppt = unit_registry("ppt")
+    assert ppt.to("ppb").magnitude == 1 / 1000
+
+
+def test_short_definition():
+    tC = unit_registry("tC")
+    np.testing.assert_allclose(tC.to("tCO2").magnitude, 44.0 / 12.0)
```
rgieseke:
```suggestion
    np.testing.assert_allclose(tC.to("tCO2").magnitude, 44 / 12)
```

tests/unit/test_units.py
znicholls: @swillner explanation here is that Python3 does floating point division by default

openscm/units.py
znicholls: I might use it

rgieseke: Fair enough!

rgieseke:
In `units.py` symbol and value are exposed at the module level, which they probably shouldn't. Not sure if there is a smart pattern to make this kind of setup code hidden, likely easiest to rename them to `_symbol` and `_value` 

```python
In [2]: openscm.units.value                                                                                                                                                                                 
Out[2]: 'SF6'
```

dev-requirements.txt
znicholls: not in a notebook

tests/unit/test_units.py
swillner: yeah, I used explicit `round` somewhere where I thougt it's necessery... There is also an `assertAlmostEqual` in `unittest` standard package. Should we agree on a standard here?

znicholls:
np.testing.assert_allclose

dev-requirements.txt
swillner: Ah, ok. We could also easily make such a function ourselves... But I don't care as long as we can rely on `pypi` as soon as we release ;)
We then would also need to pinthe right version (e.g. with `Pint` the current release does not seem to properly override existing units...)

swillner:
@rgieseke Have you disabled the "normal" merge option?

dev-requirements.txt
znicholls: sounds good

znicholls:
> Have you disabled the "normal" merge option?

yep merge commits are annoying, I prefer rebase...

swillner:
ah, ok. will you have the honor then? ;)

znicholls:
Nah in general I think authors shouldn't merge, for you or robert

PR #40, opened 2018-11-04
Low-level: Parameter handling

swillner:
- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+
+    _children: Dict[str, "_Parameter"]
+    """Child parameters"""
+
+    _data: Any
+    """Data"""
+
+    _has_been_aggregated: bool
+    """
+    Tells if parameter has already been read in an aggregated way, i.e., aggregating
```
znicholls:
```suggestion
    If True, parameter has already been read in an aggregated way, i.e., aggregating
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+    over child parameters
+    """
+
+    _has_been_written_to: bool
+    """Tells if parameter data has already been changed"""
```
znicholls:
```suggestion
    """If True, parameter data has already been changed"""
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+    Parameter type.
+    """
+
+    SCALAR = 1
+    TIMESERIES = 2
+
+
+class _Parameter:
+    """
+    Represents a parameter in the parameter hierarchy.
```
znicholls: We need to add some explanation of what we mean by the term 'Parameter' here

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+        name
+            Name
+        unit
+            Unit
+        parameter_type
+            Parameter type
+
+        Raises
+        ------
+        RegionAggregatedError
```
znicholls:
```suggestion
        ParameterAggregatedError
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+            If the subregion would need to be added and a parameter of this region has
```
znicholls:
```suggestion
            If the child would need to be added and this parameter has
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+            already been read in an aggregated way. In this case a subregion cannot be
```
znicholls:
```suggestion
            already been read in an aggregated way. In this case a child parameter cannot be
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+        ----------
+        parameter_type
+            Parameter type to be written
+
+        Raises
+        ------
+        ParameterReadonlyError
+            If parameter is read-only because it has child parameters.
+        """
+        if len(self._children) > 0:
```
znicholls:
```suggestion
        if self._children:
```

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+    """
+    Represents a region in the region hierarchy.
+    """
+
+    _children: Dict[str, "_Region"]
+    """Subregions"""
+
+    _has_been_aggregated: bool
+    """
+    Tells if a parameter of this region has already been read in an aggregated way,
```
znicholls: see comment above

openscm/core.py
```diff
@@ -1,24 +1,304 @@
...
+    _name: str
+    """Name"""
+
+    _parameters: Dict[str, _Parameter]
+    """Parameters"""
+
+    _parent: "_Region"
+    """Parent region (or `None` if root region)"""
+
+    def __init__(self, name: str):
```
znicholls: can we somehow avoid duplicating `__init__` between `Region` and `Parameter`? Superclass?

tests/unit/test_core.py
```diff
@@ -0,0 +1,85 @@
...
+    ParameterReadonlyError,
+    ParameterSet,
+    ParameterType,
+    ParameterTypeError,
+    ParameterWrittenError,
+    RegionAggregatedError,
+)
+import pytest
+
+parameterset = ParameterSet()
```
znicholls: turn this into a fixture so it's regenerated before each test

tests/unit/test_core.py
```diff
@@ -0,0 +1,85 @@
...
+
+    region_deu.attempt_aggregate()
+    with pytest.raises(RegionAggregatedError):
+        parameterset._get_or_create_region(("DEU", "BRB"))
+
+
+def test_parameter():
+    region_ber = parameterset._get_or_create_region(("DEU", "BER"))
+
+    with pytest.raises(ValueError):
```
znicholls: what message do we expect here?

openscm/core.py
swillner: Good point. I have a section in the docs for the low-level API. Repeat here?

openscm/core.py
swillner: We could in principle, but then the typing of `_parent` and `_children` would not be really correct (we would need generic programming abilities here...). Let's use these for now and the refactor once the whole thing works, ok?

tests/unit/test_core.py
swillner: What do you mean?

openscm/core.py
swillner: This is actually a different exception than `ParameterAggregatedError`, but we could reuse it also for regions. See also below.

openscm/core.py
swillner: Which one? ;)

openscm/core.py
swillner: Hmm, I would prefer to call "child regions" "subregions", but I am not strict about this. But it should read "a parameter of this region", because a region does not represent a parameter, but collects all of those for the specific region...

openscm/core.py
swillner: Either "subregion" or "child region", not "parameter"

openscm/core.py
znicholls: reference here (you can cross reference in docstrings)

tests/unit/test_core.py
swillner: Right now, there is no message, should I include one such as "No region name given"?

openscm/core.py
znicholls: hang on, in this bit aren't we looking at a call to `get_or_create_child_parameter`, there are no regions anywhere (not in the function call, nor in the implementation)?

openscm/core.py
znicholls: good plan. we may need to use different words other than parent and child given their pre-existing special meanings

tests/unit/test_core.py
znicholls: parameterset should be a pytest fixture so we get a fresh instance of ParameterSet for each test. The way you have it set up, the same instance of ParameterSet is used in all of the tests and hence we could get some weird fails because something that was set in one test blows things up in another (when it shouldn't).

openscm/core.py
znicholls: changing 'Tells if a' to 'If True, a'

tests/unit/test_core.py
znicholls: yes please, you can test with 

```python
error_msg = re.escape("Expected error msg")
with pytest.raises(ValueError, match=error_msg):
   code to run
```

openscm/core.py
swillner: done

openscm/core.py
swillner: sorry, my bad, didn't see the context. done

openscm/core.py
swillner: done

tests/unit/test_core.py
swillner: done

tests/unit/test_core.py
swillner: done

openscm/core.py
swillner: I also missed the context here, sorry. done

openscm/core.py
swillner: sorry, done ;)

swillner:
@znicholls thanks for the feedback! further one is only needed after I have continued the work here...

openscm/core.py
swillner: As I perceive it, these words are rather common when programming hierarchical structures. I would, personally like to stick to them.

znicholls:
@swillner how far away are we on this? Would be brilliant to get into so I can have a crack and putting the full pipeline together with the PH99 adapter

swillner:
@znicholls I think I can finish it next week (this week on vacation). ok?

znicholls:
Looking good @swillner, if you re-run the notebooks locally you'll pass CI

openscm/parameters.py
```diff
@@ -0,0 +1,264 @@
...
+from .timeframes import Timeframe
+
+
+class ParameterType(Enum):
+    """
+    Parameter type.
+    """
+
+    SCALAR = 1
+    TIMESERIES = 2
```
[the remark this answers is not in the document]
swillner: `ParameterType` is an enum, so we don't have to deal with these numbers anyway, e.g. using `ParameterType.SCALAR`.

docs/conf.py
```diff
@@ -176,4 +182,14 @@
 ]
 
 
+def skip(app, what, name, obj, skip, options):
+    if name == "__init__":
+        return False
+    return skip
```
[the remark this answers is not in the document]
swillner: I clarified that now: just renamed the function `skip` to `skip_init`

swillner:
Great! I still need to include aggregated parameter views (#69), but I think we can merge this for now...

openscm/parameters.py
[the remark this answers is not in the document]
znicholls: so can we just get rid of those lines? I'm guessing the answer is no, I just started having nightmares of using integers as flags when I saw these lines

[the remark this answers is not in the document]
znicholls: ok, https://docs.python.org/3/library/enum.html, nightmares gone

znicholls:
@rgieseke for you to hit the button

Issue #22, opened 2018-10-12
Define standard parameter hierarchy


swillner:
Parameters (scalar values or time series) are organized in a hierarchy. Independent from the implementation itself, OpenSCM defines a standard hierarchy that its model adapters should adhere to and end-users can rely on.

In documentation and probably in the standard file format, the hierarchical levels are separated by |, e.g., “Emissions|CO2|Land use”.

swillner:
Please also mention example units for each parameter (OpenSCM should be agnostic about the specific unit) to collect which units should be supported by the unit converter, #23

swillner:
Reminder: Especially define which *type of unit* each parameter expects and which baseline it shoudl refer to if applicable (see #45)

znicholls:
Bits of the parameter space to think about (copied from #96)

- gas species
- sectors
- regions/spatial dimensions

swillner:
Also define which parameters should be read in `Adapter._initialize_run_parameters` and which in `Adapter._initialize_model_input` (see #104)

From #104, @znicholls :

> I can't think of the right words to use to make the distinction between scenario specific stuff (e.g. emissions) and run specific stuff (e.g. ECS). I'm not sure that 'run parameters' and 'model input' is very clear. @rgieseke @chrisroadmap how would you distinguish between stuff that changes every run and stuff the changes once every, say, 600 runs when doing an ensemble of runs?

PR #41, opened 2018-11-04
Adapter: DICE climate component

swillner:
- [ ] Rebase after merge of #146
- [ ] Rename parameters (#22) and add to #139 
- [ ] Simplify parameter view usage
- [ ] In-code TODOs
- [x] Tests added
- [x] Documentation added


openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+        self._timestep = 0
+        self._timestep_count = (
+            self._stop_time - self._start_time
+        ) // self._period_length
+        self._year2100 = datetime(2100, 1, 1).timestamp()
+
+        # Read model parameters
+        model_parameter_defaults = {
+            # Initial concentration in atmosphere 2010 (GtC)
+            "M_atm0": (830.4, "GtC/a"),
```
znicholls: Is there a way to avoid writing the defaults in two places (as is currently effectively the case with this dictionary plus the list above)?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+    _year2100: int
+    """Timestamp of 2100-01-01"""
+
+    @abstractmethod
+    def initialize(self) -> None:
+        """
+        Initialize the model.
+        """
+
+        self._period_length = constants.YEAR
```
znicholls: make constants local please

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+            "c3": (0.09175, "W/m^2/°C"),
+            # Transfer coefficient for lower level
+            "c4": (0.00487, ""),
+            # Forcings of equilibrium CO2 doubling (Wm-2)
+            "fco22x": (3.8, "W/m^2"),
+            # 2010 forcings of non-CO2 GHG (Wm-2)
+            "fex0": (0.25, "W/m^2"),
+            # 2100 forcings of non-CO2 GHG (Wm-2)
+            "fex1": (0.7, "W/m^2"),
+            # Equilibrium temp impact (°C per doubling CO2)
```
znicholls:
```suggestion
            # Equilibrium temp impact (aka ECS) (°C per doubling CO2)
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+        self._b33 = 1 - self._b32
+
+        # Total CO2 emissions (GtCO2 per year)
+        self._E = self._parameterset.get_writable_timeseries_view(
+            ("Emissions", "CO2"), (), "GtCO2/a", self._start_time, self._period_length
+        )
+
+        # Concentration in atmosphere (GtC)
+        self._M_atm = self._parameterset.get_writable_timeseries_view(
+            ("Concentration", "Atmosphere"),
```
znicholls:
```suggestion
            ("Pool", "CO2", "Atmosphere"),
```

'Pool' isn't the greatest term but might be ok (perhaps 'change in burden' throughout would be better?). In any case, the units give away that this isn't a concentration.

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+            (),
+            "GtC",
+            self._start_time,
+            self._period_length,
+        )
+        self._M_l.set(0, self._M_l0)
+
+        # Carbon concentration increase in shallow oceans (GtC from 1750)
+        self._M_u = self._parameterset.get_writable_timeseries_view(
+            ("Concentration", "Ocean", "shallow"),
```
znicholls:
```suggestion
            ("Pool", "CO2", "Ocean", "shallow"),
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+            (),
+            "GtC",
+            self._start_time,
+            self._period_length,
+        )
+        self._M_u.set(0, self._M_u0)
+
+        # Increase in temperature of lower oceans (°C from 1900)
+        self._T_ocean = self._parameterset.get_writable_timeseries_view(
+            ("Temperature Increase", "Ocean", "lower"),
```
znicholls:
```suggestion
            ("Ocean Temperature Anomaly", "lower"),
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+            (),
+            "GtC",
+            self._start_time,
+            self._period_length,
+        )
+        self._M_atm.set(0, self._M_atm0)
+
+        # Carbon concentration increase in lower oceans (GtC from 1750)
+        self._M_l = self._parameterset.get_writable_timeseries_view(
+            ("Concentration", "Ocean", "lower"),
```
znicholls:
```suggestion
            ("Pool", "CO2", "Ocean", "lower"),
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+        self._T_ocean.set(0, self._T_ocean0)
+
+        # Increase in radiative forcing (watts per m2 from 1900)
+        self._force = self._parameterset.get_writable_timeseries_view(
+            ("Radiative forcing",), (), "W/m^2", self._start_time, self._period_length
+        )
+
+        # Increase temperature of atmosphere (°C from 1900)
+        self._T_atm = self._parameterset.get_writable_timeseries_view(
+            ("Temperature Increase", "Atmosphere"),
```
znicholls:
```suggestion
            ("Surface Temperature Anomaly"),
```

There's almost no way this is meant to be anything other than surface temperature.

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+            self.step()
+
+    def step(self) -> None:
+        """
+        Do a single time step.
+        """
+
+        self._timestep += 1
+
+        # Concentration in atmosphere (GtC)
```
znicholls:
```suggestion
        # Atmospheric pool size
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+        this_M_atm = max(
+            self._M_atm_lower,
+            last_M_atm * self._b11
+            + last_M_u * self._b21
+            + E[self._timestep - 1] * self._period_length / constants.YEAR / 3.666,
+        )
+
+        # Carbon concentration increase in lower oceans (GtC from 1750)
```
znicholls:
```suggestion
        # Lower ocean pool size
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,303 @@
...
+        this_M_l = max(self._M_l_lower, last_M_l * self._b33 + last_M_u * self._b23)
+
+        # Carbon concentration increase in shallow oceans (GtC from 1750)
```
znicholls:
```suggestion
        # Shallow ocean pool size
```

openscm/adapters/dice.py
rgieseke: Not a fan of "anomaly", see page 27 in https://coast.noaa.gov/data/digitalcoast/pdf/psychology-climate-change-communication.pdf

"Surface Temperature Increase"?

openscm/adapters/dice.py
swillner: Sorry, I requested the review too soon. Will do

openscm/adapters/dice.py
swillner: have already, PR is not up to date, sorry

openscm/adapters/dice.py
swillner: Waiting for #22

openscm/adapters/dice.py
swillner: Will change these according to #22 once we have that

openscm/adapters/dice.py
swillner: These are the comments from the original DICE source. Shall we stick to these, though "concentration" is not correct here?

swillner:
- [ ] Rebase once #127 is merged

openscm/adapters/dice.py
znicholls: I'd prefer to copy the DICE notation, but fix the comments to actually represent what they are (I'm not a fan of comments which are wrong but am ok with using misleading notation to maintain consistency)

openscm/adapters/dice.py
znicholls: ok but our audience here is scientists no (I can't imagine using the source as a communication tool...?)

openscm/adapters/dice.py
swillner: Cool, will mark them

openscm/adapters/dice.py
swillner: > ok but our audience here is scientists no (I can't imagine using the source as a communication tool...?)

What do you mean by that?
All adapters should use the standard names which we are defining in #22, so let's do that first

znicholls: (maybe this belongs in #22) @rgieseke and I are disagreeing about what the standard name for 'Surface Temperature Change' should be. @rgieseke links https://coast.noaa.gov/data/digitalcoast/pdf/psychology-climate-change-communication.pdf and hence against using 'anomaly', I'd argue that our audience (also referenced in Section 1 or something of that doc) is not the public but rather scientists so let's use the word that means what we mean and stick to 'anomaly'.

swillner: ah, ok ;)

swillner:
This should be working after merging of and then rebasing to #144 and #145.
Missing:
- [ ] Some minor in-code TODOs

znicholls:
Ok awesome I'll have a look early next week

swillner:
@znicholls
- No hurry ;) I won’t be doing much before April. Thinking about making the use of views easier, though, which would also simplify the code here (joining the `_values` and `_views` attributes).
- How shall we deal with required parameters in the tests (emissions in this case, see TODO in code)?

CHANGELOG.rst
```diff
@@ -17,5 +17,6 @@ master
 - (`#83 <https://github.com/openclimatedata/openscm/pull/83>`_) Rename OpenSCMDataFrame to ScmDataFrame
 - (`#78 <https://github.com/openclimatedata/openscm/pull/78>`_) Added OpenSCMDataFrame
 - (`#44 <https://github.com/openclimatedata/openscm/pull/44>`_) Add timeframes module
+- (`#41 <https://github.com/openclimatedata/openscm/pull/41>`_) Add DICE adapter
```
znicholls: Do we want to put this in order of time rather than PR number?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+from ..timeseries_converter import (
+    ExtrapolationType,
+    InterpolationType,
+    create_time_points,
+)
+
+YEAR = 365 * 24 * 60 * 60
+
+MODEL_PARAMETER_DEFAULTS = {
+    # Initial concentration in atmosphere
```
znicholls:
```suggestion
    # Initial size of carbon pool in atmosphere (referred to as concentration in original publication but units show this is wrong
```

znicholls: same for all the pools below which have units of "GtC"

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+    "M_atm0": (830.4, "GtC"),
+    # Equilibrium concentration atmosphere
+    "M_atm_eq": (588, "GtC"),
+    # Lower bound concentration in atmosphere
+    "M_atm_lower": (10, "GtC"),
+    # Initial concentration in lower strata
```
znicholls:
```suggestion
    # Initial concentration in lower stratosphere
```

znicholls: similarly below, I think using shortenings here will just be confusing

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+    Adapter for the climate component from the Dynamic Integrated Climate-Economy (DICE)
+    model.
+
+    TODO recalculate increase to absolute values
+    TODO still calibrated to initial year 2010
+    TODO What about timeseries parameters when run times change?
+    """
+
+    _period_length: int
+    """Period length in seconds (currently 1yr)"""
```
znicholls:
```suggestion
    """Period length in seconds (default 1yr)"""
```

?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+
+    def _initialize_model_input(self) -> None:
+        pass
+
+    def _initialize_run_parameters(self) -> None:
+        self._timestep = 0
+        self._timestep_count = (
+            self._stop_time - self._start_time
+        ) // self._period_length
```
znicholls: maybe +1 here so it runs over the stop_time rather than stopping before it (it's easy to filter out an extra timestep after if you really don't want it but running again can be annnoying)?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+                "GtCO2/a",
+                time_points_for_averages,
+                ParameterType.AVERAGE_TIMESERIES,
+                InterpolationType.LINEAR,
+                ExtrapolationType.LINEAR,
+            ).set(np.zeros(self._timestep_count))
+
+        # Concentration in atmosphere
+        self._views.M_atm = self._output.get_writable_timeseries_view(
+            ("Concentration", "Atmosphere"),
```
znicholls:
```suggestion
            ("Atmospheric Concentrations", "CO2"),
```

following #139

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+            ("World",),
+            "GtC",
```
znicholls: following above comments, this unit makes no sense for atmospheric concentrations (you can't convert GtC into ppm with our current unit_registry) so I think we have to rename this variable

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+            time_points,
+            ParameterType.POINT_TIMESERIES,
+            InterpolationType.LINEAR,
+            ExtrapolationType.LINEAR,
+        )
+
+        # Increase temperature of atmosphere (rel. to 1900)
+        self._views.T_atm = self._output.get_writable_timeseries_view(
+            ("Temperature Increase", "Atmosphere"),
```
znicholls: alter to follow #139

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+            ("Temperature Increase", "Ocean", "lower"),
+            ("World",),
+            "degC",
+            time_points,
+            ParameterType.POINT_TIMESERIES,
+            InterpolationType.LINEAR,
+            ExtrapolationType.LINEAR,
+        )
+
+        # Increase in radiative forcing (rel. to 1900)
```
znicholls: is radiative forcing actually relative in DICE?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+            time_points,
+            ParameterType.POINT_TIMESERIES,
+            InterpolationType.LINEAR,
+            ExtrapolationType.LINEAR,
+        )
+
+    def _reset(self) -> None:
+        self._timestep = 0
+        for name in MODEL_PARAMETER_DEFAULTS:
+            setattr(self._values, name, getattr(self._views, name).get())
```
znicholls: so we reset to whatever's in the views, not to zero/defaults?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+        self._values.M_l = np.empty(self._timestep_count)
+        self._values.M_l[0] = self._values.M_l0
+        self._values.M_u = np.empty(self._timestep_count)
+        self._values.M_u[0] = self._values.M_u0
+        self._values.T_atm = np.empty(self._timestep_count)
+        self._values.T_atm[0] = self._values.T_atm0
+        self._values.T_ocean = np.empty(self._timestep_count)
+        self._values.T_ocean[0] = self._values.T_ocean0
+
+        self._values.force = np.empty(self._timestep_count)
```
znicholls:
```suggestion
        self._values.rf = np.empty(self._timestep_count)
```

or `forcing` or `radiative_forcing`, `force` is not a commonly used abbreviation

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+        """
+        self._timestep += 1
+        self._current_time += YEAR
+
+        # Atmospheric pool size
+        self._values.M_atm[self._timestep] = max(
+            self._values.M_atm_lower,
+            self._values.M_atm[self._timestep - 1] * self._values.b11
+            + self._values.M_u[self._timestep - 1] * self._values.b21
+            + self._values.E[self._timestep - 1] * self._period_length / YEAR / 3.666,
```
znicholls:
```suggestion
            + self._values.E[self._timestep - 1] * self._period_length / YEAR / 44 / 12,
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+        )
+
+        # Lower ocean pool size (rel. to 1750)
+        self._values.M_l[self._timestep] = max(
+            self._values.M_l_lower,
+            self._values.M_l[self._timestep - 1] * self._values.b33
+            + self._values.M_u[self._timestep - 1] * self._values.b23,
+        )
+
+        # Shallow ocean pool size (rel. to 1750)
```
znicholls: is this actually relative to 1750 or is it relative to eqm?

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+                    - self._values.T_ocean[self._timestep - 1]
+                ),
+            ),
+        )
+
+        # Exogenous forcing for other greenhouse gases
+        if self._start_time + self._period_length * self._timestep > self._year2100:
+            forcoth = self._values.fex1
+        else:
+            forcoth = (
```
znicholls: is this a linear interpolation or a weighted mean or something?

tests/adapters/test_dice.py
```diff
@@ -0,0 +1,59 @@
...
+
+from openscm.adapters.dice import DICE, YEAR
+from openscm.parameters import ParameterType
+from openscm.timeseries_converter import create_time_points
+
+
+def _run_and_compare(test_adapter, filename):
+    original_data = pd.read_csv(filename)
+    start_time = int(datetime(2010, 1, 1).timestamp())
+    timestep_count = 100
```
znicholls: why so few?

CHANGELOG.rst
swillner: To be honest I don't believe this list makes sense before the first release. But by number is good enough as the order per release section, I think. Otherwise, merging messes up the order easily, anyway.

openscm/adapters/dice.py
swillner: Ok. Will rename the comments and parameters together with a PR to #139

openscm/adapters/dice.py
swillner: Oh, good point. Sorry, missed that we said including the stop time. Will also add a base test for that.

swillner:
Agreed. Will alter the names last (see above) along with a PR to #139

swillner:
What do you mean with your comment about the base tests? Isn't that for a different PR?
What about adding a function to set the required parameters that has to be implemented by the respective adapter test?

openscm/adapters/dice.py
znicholls: As in the period length could be set by the user so 'currently' doesn't really make any sense. It is by default 1yr, but if the user really forced things, they could change it. Maybe deleting 'currently' would also be fine, but the word 'currently' here implies that it might have been something else in the past/future so is just confusing.

CHANGELOG.rst
znicholls: > Otherwise, merging messes up the order easily, anyway

Not if you're careful but it really doesn't matter

znicholls:
> What do you mean with your comment about the base tests? Isn't that for a different PR?

There's a whole bunch of TODOs in there that should be done before we merge this. Perhaps in a different PR but it's just a reminder for us I guess.

> What about adding a function to set the required parameters that has to be implemented by the respective adapter test?

Sounds useful.

openscm/adapters/dice.py
swillner: as said above these are the original names (you should also the economic variables) and i will change them once we have a standard

openscm/adapters/dice.py
swillner: yes

openscm/adapters/dice.py
swillner: yes, see description of the reset function

openscm/adapters/dice.py
swillner: will keep the original names (though i also dont agree with most choices). will add acomment about that

openscm/adapters/dice.py
swillner: again, part of the original code. will have to look into the refernced papers.

openscm/adapters/dice.py
swillner: linear interpolation. all code is from the original!

openscm/adapters/dice.py
swillner: i have a list of all i need to alter, no worries

openscm/adapters/dice.py
swillner: please no further suggestions here, have already started on these, see other comments

openscm/adapters/dice.py
swillner: (keeping open as areminder to myself)

openscm/adapters/dice.py
swillner: yes, it’s meant to be changable in the future. will add a TODO instead

openscm/adapters/dice.py
swillner: hmm, what about putting pur comments and then only quote the original code comments? would want to have them in there for people to easily compare to the original dice code

swillner:
Sorry, I am well aware many of the names etc. are not the best and in many cases just incorrect. Will try to include both, improved naming and original ones.

znicholls:
>     * No hurry ;) I won’t be doing much before April.

Sorry only just saw this

>     * Thinking about making the use of views easier, though, which would also simplify the code here (joining the `_values` and `_views` attributes).

sounds gewd

>     * How shall we deal with required parameters in the tests (emissions in this case, see TODO in code)?

Something like https://github.com/openclimatedata/openscm/blob/bfa715bfc468bd31b0aa435f2017403aed7739a7/tests/integration/test_adapters.py#L36 ?

openscm/adapters/dice.py
znicholls: ah ok that's sensible then. I'm wondering if we should have a disclaimer at the top that says something like, "In the following we stick as closely as possible to the original dice code. We then use the adapter to translate the DICE conventions into OpenSCM conventions, effectively resolving any confusing DICE internal conventions in the process for any OpenSCM users."

openscm/adapters/dice.py
znicholls: ok an in future 1yr would be the default value?

openscm/adapters/dice.py
znicholls: fair, goodness me though these comments are definitely not making life easy

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+                self._values.fex0
+                + (self._values.fex1 - self._values.fex0)
+                * (self._period_length / YEAR * 0.2 * self._timestep)
+                / 18.0
```
znicholls:
```suggestion
                / 90.0
```

openscm/adapters/dice.py
```diff
@@ -0,0 +1,388 @@
...
+        )
+
+        # Exogenous forcing for other greenhouse gases
+        if self._start_time + self._period_length * self._timestep > self._year2100:
+            forcoth = self._values.fex1
+        else:
+            forcoth = (
+                self._values.fex0
+                + (self._values.fex1 - self._values.fex0)
+                * (self._period_length / YEAR * 0.2 * self._timestep)
```
znicholls:
```suggestion
                * (self._period_length / YEAR * self._timestep)
```

openscm/adapters/dice.py
swillner: yes

swillner:
@znicholls still working on this... will give you a ping, hwen it's done for another review ;)

PR #44, opened 2018-11-05
Low-level: Timeframe conversion

swillner:
- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


openscm/timeframes.py
```diff
@@ -0,0 +1,65 @@
+class TimeframeConverter:
+    """
+    Converts timeseries and their points between two timeframes (each defined by a time
+    of the first point and a period length).
```
znicholls: Do we need the number of time periods/end point too? If no, I'm not clear what this does.

swillner: We don't necessarily need the end points. The number of points/time periods would be used from the list/array in the `convert_X` function.

swillner:
The task of the `TimeframeConverter` is to shift and scale in time according to which period in time the data points belong to.
We also need to discuss how to convert e.g. from a year (starting 1.1. 00:00:00) to months (also starting there). Should the values all be the same as for the year (since we see all of them as averages) or should we interpolate including information about the values the year before and the year after? (clear what I mean?)

openscm/timeframes.py
znicholls: got it!

znicholls:
> clear what I mean?

I think so. If we have data starting 1.1.1 00:00:00 and another data point at 2.1.1 00:00:00 and we want to go to months, I think the default should be to interpolate, it would be ideal to be able to also return an array where all values in year 1 are the same as the data in 1.1.1 00:00:00 if possible.

swillner:
so 2.1.1 is the first day of year 2, right? then yes ;)

swillner:
so 2.1.1 is the first day of year 2, right? then yes ;)

swillner:
I propose a conversion algorithm explained in `notebooks/timeframes.ipynb`

znicholls:
> I propose a conversion algorithm explained in `notebooks/timeframes.ipynb`

Definitely a bit confusing to get your head around but super nice. To make things a bit clearer, can I suggest we tweak the notebook to start with an easier example before moving onto the more complex example you have.

I propose:

1. Start with a timeseries which is for point values (i.e. start with the orange dots on the 'interpolation' line)
1. Show how you would interpolate these point values (join them all with a straight line)
1. Show how we then get a different set of points by specifying a new target_start_time and target_period_length (so the green dots in 'target edges')
1. Show the output points (they also lie on the interpolation line)
1. Show that, if we convert back to source_start_time and source_period_length, we get back our original points

Then have a cell which explains that it's all easy when you're looking interpolating points i.e. considering state variables, but that it gets much more complicated when you want to interpolate flux variables.

1. Put what you have in the notebook now here
1. Add one more plot that shows that if we directly revert to source_start_time and source_period_length (without doing any conversions and risking losing data), we get back our original line (sanity check that going round the entire loop takes us back to where we started) i.e.

```python
target_values = timeframes._convert(
    source_values, 
    source_start_time, 
    source_period_length, 
    source_start_time, 
    source_period_length
)

np.testing.assert_arrayequal(target_values, source_values)
```

swillner:
cool, will do! thanks for the quick feedback!

notebooks/timeframes.ipynb
```diff
@@ -0,0 +1,729 @@
...
...AN90wbRnFVA58eyNfaFKuSYuX+JI6E01WVdMSZ8lq+TD/DoqkRBHg7j9vzYucRGdyUr586ZCpSSn63/gArNh0nNNiHm8YPvmibqnoDi97cQ2lNA+/fP4UhffzsbbbCjiSN6EvUgEBWbc/klomDdd/ewVxSM0roa4P8tiPgPBGsmeG9h7e7/K2d2UgpWTA1wu7atuZ8XSEEf7sljoKKeh778CADAn2Yel4DrUajiQfX7ePY2UreWJDQo8oarfzVUtsRfRZCsHhaBI9/lE5KRglTh1vWUE2PPkspSckoYfKwvjb5INLSZ3MQXeVChRA+wDbAm6ZA/6GU8qnOtklISJBpafrprFXTYGDKc98xdXhfXrtrotbm2IXy2kZ++XoK+WW1/Gr6cNal5pBfVouvlzs1DUaevzmW23UwTZvCPtQ1Gpn4l28wmiT1BhOhwb4smRXlMH162pJdUs3lK7bylxujmT8lQmtzrIIQYq+UMsGcdc3JcdcDV0op44B44BohxOSeGGhLqhqqqGqoumDZR/vyKK9ttErPbUu17UF7ukG+nqxZlIgQ8Levj5NXVosEahqMeLgJqzyI1MpfLbUd1eevD52l3mCizmBC0jTxwNIN6Wzcn2dT3Z7Skbat55fU0mdzMGeyYCmlbPHAs/lHt51rDhYc5GDBwdbXJpNkzfZMxg0O6nIIt7W17UVHuqHBvu0GaINJsmLTcZvp2gO9HWu9a6/YdBxDm4ZTtY1Gs64DPfqccqqEfgFeDA+xTX5bS5/NwayqEiGEuxDiAFAIfCOl3NXOOslCiDQhRFpRUZG17ew2P5woIqO4mnum6b/nti0oqmx/4EW+FWdIUeifjs63I14HUkpSM85xiY3y246AWYFbSmmUUsYDg4FJQoiYdtZZKaVMkFImhISEWNvObrNqeyYDenlzXewgrU3RhNBgX4uWK5yTjs53gI8HtQ2OVd+dXVLD2Yo6q80v6YhYVMctpSwDvgeusY051uXY2Qq2/1zM3VMi8HTR2VuWzIrCt026xNfTnSWzojSySKEF7V0HbgIq6wxc+d9b+eRAnlmDtvRAio36kzgSXUYzIUSIECK4+W9f4GrgmK0NswZrtmfh4+nGvEtct3pi7vgwnrs5lrBgXwQQFuzLczfHOmw1gaJ7tHcd/M8v41mfPJm+AV48/N4BbvrnTvbllGptapekZpQQEu
```
znicholls: Here are we:

- taking the average of the points in the time period?

or

- actually doing the integration of the interpolated time series over the time period of interest?

If they're the same then maybe we can skip ever working out the target timeseries point values?

znicholls: 'i.a.' --> 'i.e.' (suggestions aren't working here...)

znicholls: 'not necessarily leads' --> 'doesn't necessarily lead'

znicholls: 'might lose some information' is this because we chop out some information at the start/end? Intuitively I feel like we should always go back to the same interpolated line but maybe I'm just wrong and need to do actually do the maths?

znicholls:
> Can you have a look at the updated notebook, please?

Super nice, few minors

notebooks/timeframes.ipynb
```diff
@@ -0,0 +1,729 @@
...
+   "metadata": {},
+   "source": [
+    "## Target timeframe"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A conversion of these data into a target timeframe can of course only be done for the x-values where data is available."
```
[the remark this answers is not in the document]
swillner: I am not really comfortable to do so. How would we extrapolate? Maybe shift to a later release if such a feature turns out necessary?

notebooks/timeframes.ipynb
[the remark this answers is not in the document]
swillner: done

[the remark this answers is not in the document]
swillner: Oh, I thought we agreed on treating all variables as averages?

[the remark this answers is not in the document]
swillner: ok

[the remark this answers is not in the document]
swillner: Changed it to "In a second step, the linearized values are used to determine the values at the edges of the target timeseries as well as at their inflection points."
Is that clearer?

swillner: We do the second. Changed the text accordingly.

swillner: changed to "e.g."

swillner: Done
Added: "This is because of possibly larger period lengths, overlapping of periods (both leading to averages taken over several data points possibly losing their information), and chopped-off information at the beginning and the end of the timeseries for shifted periods."

notebooks/timeframes.ipynb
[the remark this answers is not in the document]
znicholls: Good plan!

notebooks/timeframes.ipynb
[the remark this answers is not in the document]
znicholls: Sorry I may have lost track of context. Internally yes I think we should, but the user should be able to give us point values no?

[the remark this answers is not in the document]
znicholls: Are you using 'inflection points' in the mathematical sense of the point at which the second derivative changes sign or to mean something else? If something else, let's use different words as using something which is already defined to mean something else is confusing.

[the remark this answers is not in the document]
swillner: Hmm, we could offer a helper function for that. What do you think?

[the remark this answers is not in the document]
znicholls: perfect solution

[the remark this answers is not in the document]
swillner: I mean the points were the intermediate lines begin and end, "Stützstellen" in German...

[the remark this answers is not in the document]
swillner: Cool, I'll open an issue for that and will do after this PR.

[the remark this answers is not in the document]
znicholls: 'sampling points' :)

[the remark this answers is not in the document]
swillner: ok ;)

swillner:
@rgieseke @znicholls This is done so far from my side...

rgieseke:
Minor comment/question?

In "Linerarization" change "alleged" to "assumed" maybe?

Typo: averade => average

Otherwise looks good to me!

swillner:
@rgieseke Thanks!
@znicholls May I bother you for a final review? ;)

rgieseke:
@znicholls This looks good to me -- any blocking issues from your side or should we merge this?

openscm/timeframes.py
```diff
@@ -0,0 +1,513 @@
...
+    values
+        Timeseries values of period averages
+
+    Returns
+    -------
+    np.ndarray
+        Values of linearization (of length ``2 * len(values) + 1``)
+    """
+    edge_points = _running_mean(values, 2)
+    middle_points = 3 * values[1:-1] / 2 - (values[0:-2] + values[2:]) / 4
```
znicholls:
```suggestion
    middle_points = (4 * values[1:-1] - edge_points[:-1] - edge_points[1:]) / 2  # values = 1 / 2 * (edges_lower + middle_points) / 2 + 1 / 2 * (middle_points + edges_upper) / 2
```

openscm/timeframes.py
```diff
@@ -0,0 +1,513 @@
...
+        Object wrapping the interpolation points in time ("x-values" corresponding to
+        ``interpolation_values``) and additional information
+    interpolation_values
+        Interpolation values ("y-values")
+
+    Returns
+    -------
+    np.ndarray
+        Array of the interval/period averages
+    """
```
znicholls: Is there not a nicer way to write `int = (left_edge^2 - right_edge^2) / 2?

If not, can we add a comment like 'Simply calculating integral, taking care of shaping arrays correctly is fiddly'? Or split the calculation step and reshaping steps? np.add.reduceat is confusing, putting all the calculation steps in there too makes debugging tricky...

notebooks/timeframes.ipynb
```diff
@@ -0,0 +1,747 @@
...
...3ZnnePGXcbppju/mJliUFMGB3DL2Zut7dJ6zsyfrHCYJk3VybViCWYFbSnmgOX89Tko5V0qprjiFrnnh62N8fvAMT1w7mjlxoVqbcwG/mDCYXj4erN6h7rq1JDXjXLv5bUfA6XqV+HteVPDi9NquptuV9tqULP61LYP5k4dy/2XD7KZr9j68PbhjUjhvbM8kr6yWsGBfu2l3B72e556ScqqE+PBgfDwv7gappc/moPpxK5yKTYfP8sA7e7lqzABev2si7lacZd2a5JXVctnfvufeaZEsvW6M1ua4HOW1jYz/82YeunIkv9NJr3TVj1vhkuzLKeW37+4nbnAwr9w+XrdBGyAs2JdrYgby7905VNcbtDbH5UhryW87WP12C04XuH86+xM/nf3JpbRdTbc97aziau59K42BQT6sWpBwUXmXrXR7wj3TIqmsM/Dh3tN217YEPZ1na5FyqgQvDzfGh7ffn0RLn83B6XLc1Y3VLqftarpttUuq6lmwZjcAby2aRN8Ab7vo9pQJ4b2JHxLMmh2ZzJ88FLcuviG4+nm2JqmZJUzoIL9tS11r4XR33ArXorbByD1vpVFQUccbCxKI6Kfvh0ptuWdaJFklNXx3rFBrU1yG8tpGDudXOGyaBJzwjlvh/KxLX8fvN/2egqoihsq/IOrH8fr8BIcs67omZiCDgnxYvSOTq8YO0Nocl2B35jmkA+e3Qd1xKxyMdenrSP4smYKqAgINc6A+jgqf1RSbtmhtWrfwdHdjwdQIdp4q4Uh+hdbmuASpGSV4e7gRP8Rx+m+3xekCd5B3EEHetpmGSq/arqS7bMttM6YjAAAgAElEQVQyahpr8DdcRYDxaso9PuKc+JhlW5bZRd8WPt+RGI6vp3uXA3Jc6TzbUjs1o4QJ4b07zG/bSteaOF2qJLp/tMtpu5JuTnkOfobL6Gd4hGr3HyjzeLN1uT2whc9Bfp7cMnEw6/fk8vg1owkJbP/hqiudZ1tpl9U0cORMBY/M6Lx2W0ufzcHp7rgVzk2475X0a/wddW7pFHv+HUTTALLwoHCNLesZi5IiaDCaeCc1W2tTnJqW/LZeetd0F6cL3PvO7GPfmX0upe0quicLKvGpehijWwGFXs+AaBq44ufpx/IZy+1ig618HhYSwIzR/XknNZu6RqNdtbvCmf6nUprz23FDOk+DaOmzOThd4K4z1FFn
```
znicholls: Add, 'This ensures that we retain as much information as possible. We use this information in the next step, when we determine the target timeseries values'.

openscm/timeframes.py
```diff
@@ -0,0 +1,513 @@
...
+    target
+        Target timeframe
+
+    Returns
+    -------
+    _Interpolation
+        Object wrapping the actual interpolation points and additional information
+    """
+    linearization_points_len = 2 * source_len + 1
+    # Use double the time points to stick to integers when using half the source period
```
znicholls: What does this mean?

openscm/timeframes.py
```diff
@@ -0,0 +1,513 @@
...
+        ``interpolation_values``) and additional information
+    interpolation_values
+        Interpolation values ("y-values")
+
+    Returns
+    -------
+    np.ndarray
+        Array of the interval/period averages
+    """
+    interval_sums = (
```
znicholls: add comment something like interval_sums = (x2 - x1) * (y2 + y1) / 2 i.e. trapezium rule

notebooks/timeframes.ipynb
znicholls: 'Converting from source timeframe to source timeframe does not alter data - as it should.' --> 'Converting from source timeframe to source timeframe does not alter data - as we expect.'

znicholls: can we plot target (i.e. the intermediate) on here too?

znicholls:
Having said that, this is wicked thanks @swillner !!

notebooks/timeframes.ipynb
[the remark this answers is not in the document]
swillner: Didn't we decide to *only* use period average data?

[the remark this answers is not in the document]
swillner: (I also think we should avoid flux/state variables as a SCM-specific terminology that actually defines data as by their meaning rather than by how they are specified regarding timeframes)

swillner: What do you mean? This is "converting from A to A", so actually no conversion at all...

openscm/timeframes.py
swillner: I tried to clarify that in the updated comment

openscm/timeframes.py
swillner: Should be clearer now together with your suggestion about the trapezium rule

swillner:
@znicholls thanks for the review and the great suggestions. Just two open comments left and we can almost merge ;)

notebooks/timeframes.ipynb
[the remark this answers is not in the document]
znicholls: > Didn't we decide to _only_ use period average data?

Sorry yes true. We have a function to convert point data to period average data right? In that case maybe 'standard for OpenSCM' --> 'way all timeseries are internally stored in OpenSCM'.

> (I also think we should avoid flux/state variables as a SCM-specific terminology that actually defines data as by their meaning rather than by how they are specified regarding timeframes)

I don't fully understand this. The one thing I would say is that state/flux variables is not SCM-specific, it's Physics.

znicholls: Yep but it goes via the interpolation into OpenSCM's internal conventions right? Can we show what it looks like then (when it's a period average, not point data)?

swillner: ah, of course! will include that! ;)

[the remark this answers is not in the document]
swillner: You're right, sorry. I just meant, one could in principle give flux variables as point data and state variables as period averages... will include the conversion functions today ;)

PR #46, opened 2018-11-05
WIP: Basic command line interface

rgieseke:
- [ ] Tests added
- [ ] Documentation added (where applicable)
- [ ] Example added (either to an existing notebook or as a new notebook, where applicable)
- [ ] Description in ``CHANGELOG.rst`` added


tests/data/params.yaml
```diff
@@ -0,0 +1 @@
+ecs: 3
```
swillner: `.yml` instead of `.yaml` as file ending?

rgieseke: Don't care, shouldn't matter ... but:
> Is there an official extension for YAML files?
>
> Please use ".yaml" when possible.

http://yaml.org/faq.html

swillner: Oh, ok. Didn't know that. Then, let's stick to `.yaml`.

rgieseke:
`cli` folder because I expected to have more functions in sub-modules later ... could be only a single file atm.

PR #47, opened 2018-11-08
Minor fixes

swillner:


.gitignore
```diff
@@ -16,3 +16,5 @@ build/
 dist/
 docs/_build
 venv/
+.cov2emacs.log
```
rgieseke: Should this go into your global gitignore?

swillner: good point

PR #60, opened 2018-12-01
Adapter: Ph99

znicholls:
# Pull request

Add adapter for model of [Petschel-Held 1999](https://doi.org/10.1023/A:1005487123751).

- [x] #171  merged
- [x] #146 target branch
- [ ] #187 
- [ ] #186
- [x] refactored to make ph99adapter as similar to the dice adapter as possible, copy changes from #176
    - [x] `tests/unit/test_adapter.py`
    - [x] `tests/adapters/base.py`
    - [x] `tests/adapters/test_dice.py`
    - [x] `openscm/adapters/__init__.py`
    - [x] `openscm/adapters/dice.py`
    - [x] `tests/adapters/test_ph99.py`
    - [x] `openscm/adapters/ph99.py`
- [x] Considered discussions in #65
- [x] Considered discussions in #64
- [x] Tests added
- [x] Example added (either to an existing notebook or as a new notebook, where applicable) (N/A)

## Adding to CHANGELOG.rst

```
```


znicholls:
@rgieseke @chrisroadmap thoughts on this, especially #64 and #65

docs/adapters.rst
```diff
@@ -2,3 +2,5 @@ Adapters
 --------
 
 .. automodule:: openscm.adapters
+
+.. autoclass:: openscm.adapters.PH99
```
swillner: no need for this, `automodule` already takes care of the classes
```suggestion
```

docs/models.rst
```diff
@@ -0,0 +1,6 @@
+Models
+------
+
+.. automodule:: openscm.models
+
+.. autoclass:: openscm.models.PH99Model
```
swillner: same here
```suggestion
```

znicholls:
> do we actually need a separate `model` module here with all the numpy and pint overhead? Isn't an adapter enough here?

Yes an adapter is enough. However a bit like #65 I'm trying to push the scope out a bit wider to ask, "In an ideal world, without OpenSCM, how would this look?" The other thing I'm trying to get my head around is how an adapter and model should interact and starting with this simple case makes it easy, before we try and wrap our heads around how to do it with e.g. MAGICC

docs/adapters.rst
swillner: Ah, sorry, my bad. It looks like Sphinx include includes those if they are part of `__all__` for that module (or directly declared in the file). Shall we use `__all__` here or the explicit inclusion of the class?

swillner:
So, what's the best way to go here? What about having a core model class that does not care about the units and only does the actual model run and then provide an adapter inside OpenSCM and a separate layer directly accepting data?

znicholls:
> What about having a core model class that does not care about the units and only does the actual model run and then provide an adapter inside OpenSCM and a separate layer directly accepting data?

Sounds like a good option. If I get any time, I'll have a go at writing such a thing. Given that Pint arrays look and feel like numpy arrays, I might set it up so that it throws a warning about using units if the user uses numpy arrays rather than Pint arrays but otherwise it can be completely Pint/numpy agnostic.

swillner:
Any way to avoid separete model files? Either we have a model file+class for each adapter or for none. Would prefer the latter as most adapters will use a model from a separate package.

swillner:
(fyi, there will be some changes in the ParameterSet class making things easier than currently in #41)

znicholls:
> Any way to avoid separete model files?

Yes but my solution to this would be to move the PH99 implementation to a separate package. I'd like to have a 'playground' to experiment with how to set up a model which is separate from an adapter.

> Either we have a model file+class for each adapter or for none.

Why do we need this uniformity? Can't we just say that some models are maintained 'in house' whilst others must be installed from other sources?

swillner:
It’s just where I would draw the line if a model itself is to be included or in a separate package

swillner:
Cool, looking forward to see it

znicholls:
> It’s just where I would draw the line if a model itself is to be included or in a separate package

Ok shall I move PH99 to a separate package then?

znicholls:
@swillner here's another draft, hopefully gives you a sense of how this might work with an external model

openscm/utils.py
```diff
@@ -27,3 +31,32 @@ def ensure_input_is_tuple(inp: Union[str, Tuple[str, ...]]) -> Tuple[str, ...]:
...
+
+
+def convert_openscm_time_to_datetime(oscm_in: int) -> datetime.datetime:
+    """
+    Convert OpenSCM time to datetime.datetime
+    """
+    return OPENSCM_REFERENCE_TIME + relativedelta.relativedelta(seconds=oscm_in)
+
+
+def round_to_nearest_year(dtin):
```
znicholls:
```suggestion
def round_to_nearest_jan_1(dtin):
```

swillner:
@znicholls target `core-implementation`?

tests/adapters/base.py
```diff
@@ -32,23 +40,92 @@ def test_shutdown(self, test_adapter):
...
         Test that initalizing model input does as intended.
+
+        Extra tests can be adapted depending on what the adapter should actually
+        do when initialised.
         """
         assert not test_adapter._initialized
         test_adapter.initialize_model_input()
         assert test_adapter._initialized
 
+        # TODO: add tests something like below
```
swillner: What should this test? That the model sets default parameters?

openscm/adapters/__init__.py
```diff
@@ -32,6 +31,9 @@ class Adapter(metaclass=ABCMeta):
     _initialized: bool
     """``True`` if model has been initialized via :func:`_initialize_model`"""
 
+    _initialized_inputs: bool
```
swillner: I don't think we need this

tests/adapters/base.py
```diff
@@ -32,23 +40,92 @@ def test_shutdown(self, test_adapter):
...
+        # )
+        # np.testing.assert_allclose(
+        #     tadapter._parameters.get_timeseries_view(("Emissions", "CO2"), ("World",), "GtCO2/yr", expected["emissions_time_points"], ParameterType.AVERAGE_TIMESERIES).get(),
+        #     expected["emissions"],
+        #     rtol=1e-10,
+        #     atol=1e-15,
+        # )
+
     def test_run(self, test_adapter, test_run_parameters):
+        # TODO: add some tests about where input and output go to make sure adapter
```
swillner: Enough to check "Start Time" and "Stop Time" as these are the only parameters that all models will use?

tests/adapters/base.py
```diff
@@ -69,7 +148,18 @@ def test_step(self, test_adapter, test_run_parameters):
             new_time = test_adapter.step()
             assert new_time > test_run_parameters.start_time
         except NotImplementedError:
-            pass
+            pytest.skip("Step unavailable for {}".format(type(test_adapter)))
```
swillner: Nice

tests/conftest.py
```diff
@@ -431,6 +433,56 @@ def test_run_parameters():
...
+        "emissions_parameter_type": ParameterType.AVERAGE_TIMESERIES,
+        "Equilibrium Climate Sensitivity": 2.5 * _unit_registry("delta_degC"),
+        "Radiative Forcing 2xCO2": 3.5 * _unit_registry("W/m^2"),
+    }
+
+    drivers.scalar(
+        ("Equilibrium Climate Sensitivity",), "K", region=("World",)
+    ).value = setters["Equilibrium Climate Sensitivity"].magnitude
+    drivers.scalar(
+        ("Radiative Forcing 2xCO2",), "W / m^2", region=("World",)
```
swillner:
```suggestion
        ("Radiative Forcing Sensitivity",), "W / m^2", region=("World",)
```

tests/test_misc.py
```diff
@@ -0,0 +1,20 @@
+import warnings
+
+import numpy as np
+import pytest
+from conftest import assert_pint_equal
+
+from openscm.core.units import _unit_registry
+
+
+def test_pint_array_comparison():
```
swillner: Shouldn't this go into `test_units.py`?

tests/unit/test_adapter_registry.py
```diff
@@ -29,11 +29,21 @@ def test_adapter_registry_unknown_model():
         load_adapter("unknown")
 
 
-def test_adapter_registry_import_error():
+def test_adapter_registry_import_error_dice():
```
swillner: I think it totally suffices to do so for one model. It doesn't test anything about the model/adapter, only that the adapter_registry throws if dependencies cannot be found.

swillner:
Can you add a TODO that we merge model and adapter once we have a more complex adapter as for MAGICC or Hector?

znicholls:
> Can you add a TODO that we merge model and adapter once we have a more complex adapter as for MAGICC or Hector?

I'd rather not as that just feels like unnecessary work (PH99 can just sit there as a model)

openscm/adapters/__init__.py
znicholls: how else do you tell if you've initiliased inputs when making a call to `intialise_run_parameters`?

tests/adapters/base.py
znicholls: maybe nothing, I haven't worked it out yet. Current thought is that we should make this abstract and then the implementer has to decide what to test (if anything)

tests/adapters/base.py
znicholls: yes that's probably a good default unless we want to make this method abstract

tests/conftest.py
znicholls: Bad idea, `Radiative Forcing Sensitivity` is too vague (you also have e.g. concepts like `Radiative Forcing 2xCH4`)

tests/test_misc.py
znicholls: No I want to put this function in `pint` eventually but just haven't got round to it yet so am leaving it here for now (it's not OpenSCM functionality, it's a function which should be in `pint` but isn't yet)

tests/unit/test_adapter_registry.py
znicholls: I agree, I wasn't going to add this for every model but in the rebase I had it for both so just thought I'd keep them. Happy to delete if you really object but having two tests doesn't cause any harm?

swillner:
> I'd rather not as that just feels like unnecessary work (PH99 can just sit there as a model)

I thought that was the deal as a separate `model` submodule seems totally unnecessary for such simple models (and other will be external packages, anyway).

tests/conftest.py
swillner: Hmm, then we chould be consistent with `Equilibrium Climate Sensitivity`...

swillner: Have the specific gas (without "2x") in the lower level, i.e. `("Radiative Forcing Sensitivity", "CH43")`?

openscm/adapters/__init__.py
swillner: Why do we need to store that for every run? The `_initialized` is only necessary for the `OpenSCM` class to only initialize the model once it is needed for the first time. `OpenSCM` makes sure that the methods are called in the right order.

tests/test_misc.py
swillner: Ah. Does it lead to missed lines in the test coverage in OpenSCM or why do you add it here?

tests/conftest.py
znicholls: > Hmm, then we chould be consistent with `Equilibrium Climate Sensitivity`

No they're unrelated concepts. This language matches the standards in the SCM (and wider climate modelling) community.

znicholls: > Radiative Forcing Sensitivity

As a concept, this makes no sense

tests/test_misc.py
znicholls: to make sure that the custom function `assert_pint_equal`, which is in `tests/conftest.py` does what I want it to

znicholls:
> I thought that was the deal as a separate `model` submodule seems totally unnecessary for such simple models

It gives an example to other people of how we would write a stand-alone model (which is useful as it might lead to a more uniform approach to SCMs given not everyone will be happy to write within the OpenSCM framework). Modules are basically free so why are we worried about having extra ones?

tests/conftest.py
swillner: Of course they are totally different concepts (thats why it's not caleld "Temperature Sensitivity"), but how else would you call it? "Response" intead of "Sensitivity"?

tests/unit/test_adapter_registry.py
swillner: ;) No harm, I just wouldn't like to see people add that for each of their adapter ;)

swillner: Maybe choose one of the adapters as the "generic" one (in conftest) for tests that need one adapter, but don't test anything specific about it, but, for instance, the `load_adapter`?

swillner:
> It gives an example to other people of how we would write a stand-alone model (which is useful as it might lead to a more uniform approach to SCMs given not everyone will be happy to write within the OpenSCM framework). Modules are basically free so why are we worried about having extra ones?

Hmm, I wouldn't want to people to do such redundant work separating model and adapter and testing both, just bloats the project on the long run, I'm afraid. Let's have an example for a more complex model where that is actually necessary?

znicholls:
> Hmm, I wouldn't want to people to do such redundant work separating model and adapter and testing both, just bloats the project on the long run, I'm afraid.

If someone wants to write a standalone model (i.e. in a separate package because they don't want to put it in OpenSCM for whatever reason) but has no idea how to do it, where should they get inspiration?

Could we just put a comment saying something like, "For simple models, we strongly recommend only writing an adapter. If you absolutely must write a standalone model package, then we provide `ph99` as an example of how you might consider writing such a package. Of course, there are many choices to be made and other standalone model packages such as Pyhector and Pymagicc provide other examples."

openscm/adapters/__init__.py
znicholls: > The `_initialized` is only necessary for the `OpenSCM` class to only initialize the model once it is needed for the first time

Similarly, `_initialized_inputs` ensures that OpenSCM only sets the inputs which are expensive (e.g. writing input files in MAGICC) once. If `initialize_run_parameters` is public, don't we need this sort of check so users can't accidentally call `initialize_run_parameters` without calling `initialize_model_input`?

tests/conftest.py
znicholls: Like I've called it, following the convention in the community

tests/unit/test_adapter_registry.py
znicholls: I'll just get rid of it to avoid confusion. If you're confused, other implementers definitely will be!

tests/conftest.py
swillner: Hmm, that's not a proper name, just an identifier. Then just `("Radiative Forcing", "CH4")`?

znicholls: How do you mean not a proper name? 'Equilibrium Climate Sensitivity' is a quantity with a precise definition, as is 'Radiative Forcing 2xCO2', do you want it spelt out even more like 'Radiative Forcing due to Doubling of Atmospheric CO2 Concentrations'?

swillner:
I think we have different philosophies here. I would try to do abstractions and embrace these as much as possible. That includes not solving the same problem (e.g. unit conversion for which we already have an abstraction) in different ways in different places of the same project. This only leads to redundancies and decreases overall robustness. Maybe we now feel our lack of proper planing beforehand. ;)
With that being said, go ahead if @rgieseke and @lewisjared are fine with it... ;)

znicholls:
> I think we have different philosophies here

Maybe, although I'm hoping it's just a miscommunication. Let me try and restart the conversation. 

I've never been able to write a model from scratch. PH99 is a chance to do so hence I took that chance. I'd like to keep it as a model, separate from OpenSCM. Hence the options are: a) make it a new project or b) leave it in OpenSCM. I agree that the weakness of a) is that you solve the same problem twice in the same project. This might actually be a fatal weakness and a factor which makes the decision to go for b) for us. The weakness of b) is that it's slightly more mucking around and we don't have our own 'in house' model to play with. However, having thought it through now, if you want to go for a) I'll put that in an issue.

znicholls:
closing this whilst we address #174-#180

znicholls:
Superceded by #191

PR #67, opened 2018-12-13
Add Second Assessment Report GWP100 conversions

znicholls:
# Pull request

- [ ] Tests added
- [ ] Documentation added (where applicable)
- [ ] Example added (either to an existing notebook or as a new notebook, where applicable)
- [ ] Description in ``CHANGELOG.rst`` added

## Adding to CHANGELOG.rst

```
```


openscm/units.py
```diff
@@ -217,20 +237,62 @@ def _add_gases_to_unit_registry(unit_registry, gases):
 unit_registry.define("ppb = 1000 * ppt")
 unit_registry.define("ppm = 1000 * ppb")
 
+
 # Contexts:
+context_definitions = join(dirname(__file__), "metric_values.csv")
```
swillner: Let's avoid reading files on import.

znicholls: What's a good way to store tabular data then? Only read when requested?

swillner: hmm, what about a dict for each context only including gases with values? or a dict of these mapping context name to the dict with gases->values.

swillner: though additionally lazy-loading contexts would be nice

znicholls: > what about a dict for each context only including gases with values

being able to edit these values in a csv editor would be ideal given they're almost always tables of values... worst case dicts which we then write a nice wrapper to at least look at tables is fine

PR #71, opened 2019-01-13
Travis fix

swillner:
Fix testing with travis-ci


.travis.yml
```diff
@@ -3,18 +3,10 @@ cache: pip
...
-      - libboost-filesystem1.55-dev
-      - g++-4.8
+  - "3.7"
 
 before_install:
-  - pip install pip --upgrade
+  - pip install pip pytest --upgrade
 
 install:
   # annoyingly we can't put this in setup.py...
```
swillner: We should now be able to use `pint` from `pypy` as they have a new release: https://github.com/hgrecco/pint/releases

swillner: @znicholls Can we get rid of the dependency on https://github.com/thedataincubator/expectexception ? With unit tests we don't need testing for exceptions in notebooks or do we?

swillner: (failing of the current travis build seem to be related to new commits on `pint`'s master...)

rgieseke: It's for allowing cells to fail with an exception in a notebook.

swillner: Yes, but why do we need that?

swillner: (I would just like to get rid of dependencies on repositories rather than proper packages)

rgieseke: > Yes, but why do we need that?

To show to users how things can go wrong and which errors to expect?

> (I would just like to get rid of dependencies on repositories rather than proper packages)

I think the usual solution here is to vendor the code, i.e. just include it in our repo.
Since it is only needed in the notebooks you could just copy the module folder to our notebooks directory and also copy license and their Readme file into it.

Otherwise i think it's not much different having somthing on GitHub vs. PyPI (you could pin the dependency to a specific commit) - both could disappear.

swillner: Hmm, ok, I will have a look and maybe just use a `try ... except` for that purpose...

rgieseke: > Hmm, ok, I will have a look and maybe just use a try ... except for that purpose...

Which purpose? Not sure i understand.

znicholls:
Nice change. As @rgieseke pointed out the idea of using expect exception was to make the notebook run whilst giving the same output as you'd get if an error was actually thrown. I think getting rid of it whilst it's not pip installable is a good idea given that it makes the install process unnecessarily complex.

PR #73, opened 2019-01-15
Add expect exception back in

znicholls:
# Pull request

Add expect exception back in because it can go in setup.py

- [x] Tests added (N/A)
- [x] Documentation added (where applicable) (N/A)
- [x] Example added (either to an existing notebook or as a new notebook, where applicable) (N/A)
- [x] Description in ``CHANGELOG.rst`` added (N/A)

## Adding to CHANGELOG.rst

```
```


rgieseke:
To clarify, the advantage of having expect exception that you see the full exception as it would happen if a user made the same "mistake" on their machine, right?

Seems useful. I still think we could just drop the folder from the project in the notebooks directory, along with their Readme and License file.

znicholls:
> To clarify, the advantage of having expect exception that you see the full exception as it would happen if a user made the same "mistake" on their machine, right?

That's exactly it. The guys who wrote it wrote it so that they could show students errors (which are important in learning) and have their notebooks pass CI. They didn't want to use try except blocks as their students didn't know what they were and it doesn't show you the error you'd get if you used it without a try except block. 

I'm also not that fussed, I made this PR more as an illustration (to myself) of how to do this install from path in setup.py rather than requirements. Happy to tweak as @swillner suggested or close.

swillner:
Hmm. Then to keep the number of dependencies low, I strongly suggest we just do a try..except and only print the exception message (without backtrace). That way it also does not look like an error in the notebook at first glance...
fyi, I'll be on vacation the next two weeks

znicholls:
> Then to keep the number of dependencies low, I strongly suggest we just do a try..except and only print the exception message (without backtrace).

Perfect then I'll close this.

znicholls:
Another reason to not have url links as dependencies, https://github.com/pypa/pip/issues/4187 (tl;dr if you have url dependencies in setup.py, you can't upload your package to PyPI)

PR #74, opened 2019-01-16
Include expect exception

rgieseke:
This demos the vendoring of `expectexception` in the notebooks dir, as discussed in #73


rgieseke:
@swillner @znicholls Let me know what you think ...

This keeps expectexception but avoids the "install from github" which bothered @swillner i believe

swillner:
hmm, I am not very happy with either solution, especially as I do not see the actual benefit of it.
Actually, just printing the caught exception should be enough. And then it would not look like there is an error in the notebook...
In general, we should avoid pulling in dependencies unless really necessary, which is not the case here, I think. 
Having said that, I am open to be overruled ;) If there is no way around this dependency, I would include the git path, but fix the commit so we won't run into surprises as with pint.

rgieseke:
Cool, then, let's close this one and continue any discussion (if necessary) in #73

PR #94, opened 2019-02-06
Change underlying dataformat of ScmDataFrame

lewisjared:
# Pull request

Please confirm that this pull request has done the following:

- [x] Tests added
- [x] Documentation added (where applicable)
- [x] Example added (either to an existing notebook or as a new notebook, where applicable)
- [x] Description in ``CHANGELOG.rst`` added

## Adding to CHANGELOG.rst

Please add a single line in the changelog notes similar to one of the following:

```
- (`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Added feature which does something
- (`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Fixed bug identified in `#YY <https://github.com/openclimatedata/openscm/issues/YY>`_
```

PR #97, opened 2019-02-07
Fix typo in manifest in

znicholls:
@lewisjared I'll try to fix missing files here


znicholls:
@lewisjared with this #94 is good to go

PR #103, opened 2019-02-08
Add resample method for ScmDataFrame

lewisjared:
# Pull request

- [x] Tests added
- [x] Example added (either to an existing notebook or as a new notebook, where applicable)

## Adding to CHANGELOG.rst

```
- (`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Added feature which does something
- (`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Fixed bug identified in `#YY <https://github.com/openclimatedata/openscm/issues/YY>`_
```


lewisjared:
Note this currently uses master from xarray

znicholls:
> This looks like it should actually use the low-level api for consistency... What does it do differently and why is it necessary for the high-level api?

It's a sprint solution which we need for this week. Happy to discuss more when we propose merging anything into master.

swillner:
hmm, shound we have a separate branch for your urgent stuff instead of including in master and then reverting later?

znicholls:
> sorry, I should really look at what's the target branch ;)

do you want us to use a label for stuff that's not going into master?

swillner:
>> sorry, I should really look at what's the target branch ;) ...

@znicholls would be great! or use a fork for these sprints. are you going to have more like these in the coming weeks?

znicholls:
> or use a fork for these sprints

I thought about that, seemed antisocial but would happily switch if you prefer...

> are you going to have more like these in the coming weeks?

No I anticipate this is the only one.

swillner:
>> or use a fork for these sprints ...

@znicholls then just keep going as you have and I'll look at the target branch properly ;) next time let's just communicate that to the whole team. we could also just accept PRs from private forks as feature branches and use branches in the openscm repo only for big stuff...

lewisjared:
> This looks like it should actually use the low-level api for consistency... What does it do differently and why is it necessary for the high-level api?

@swillner  What do you mean by this? The hackery is to handle use cases where time series span longer than 292 years - pandas.resample doesn't handle that at all.

swillner:
> > This looks like it should actually use the low-level api for consistency... What does it do differently and why is it necessary for the high-level api?
> 
> @swillner What do you mean by this? The hackery is to handle use cases where time series span longer than 292 years - pandas.resample doesn't handle that at all.

As I (maybe wrongly) understand the purpose of this feature is to change the resolution of a time series (aka time frames as we called in the docs). We have something like that already in the "low-level API" (shoul really change that name) with the ``TimeframeConverter`` class. But maybe I just misunderstood what you are doing in this PR.

znicholls:
> As I (maybe wrongly) understand the purpose of this feature is to change the resolution of a time series (aka time frames as we called in the docs). We have something like that already in the "low-level API" (shoul really change that name) with the `TimeframeConverter` class. But maybe I just misunderstood what you are doing in this PR.

I think we're all agreeing here. We should have one way to do this, and it probably should be via the low-level API in the end, but we just haven't got to it yet. Let's revisit when we come to a merge into master.

PR #107, opened 2019-02-08
Core docs

swillner:
Fix documentation of the OpenSCM core/low-level API

PR #86, opened 2019-02-01
Simplify parameterset reading

znicholls:
# Pull request

Please confirm that this pull request has done the following:

- [x] Tests added
- [x] Documentation added (where applicable)
- [x] Example added (either to an existing notebook or as a new notebook, where applicable)
- [x] Description in ``CHANGELOG.rst`` added

## Adding to CHANGELOG.rst

Please add a single line in the changelog notes similar to one of the following:

```
- (`#XX <http://link-to-pr.com>`_) Added feature which does something
- (`#XX <http://link-to-pr.com>`_) Fixed bug identified in (`#XX <http://link-to-issue.com>`_)
```

Simplifies reading of parameterset to make it able to handle string input and also gives root a name rather than being accessed via an empty tuple. This PR also includes a wider update of the docs (cosmetic rather than content).

PR #110, opened 2019-02-08
Core docs adjustment

swillner:
Adjust documentation regarding root region.
(@znicholls Far fewer changes than I thought ;) )


znicholls:
would merging this close #107? Or just change its scope?

swillner:
Once we also merge #86, yes, I think this would close it (closing now, ok?).

znicholls:
> Once we also merge #86, yes, I think this would close it (closing now, ok?).

Ok. Whoever is second reviewer on #86 will have some reading to do then but otherwise ok.

docs/usage.rst
```diff
@@ -3,35 +3,66 @@ Usage
...
 Time frames
 -----------
 
-Timeseries parameters are always given with a corresponding time frame, which consists of a time point and a period length. The time point gives the start of the timeseries; the period length gives the length of the period between consecutive values in the timeseries. Each parameter value is assumed to be the **average value** for its corresponding period. This implies that additive values such as emissions need to be given as a rate, e.g. ``tC/a`` rather than ``tC``.
+Timeseries parameters are always given with a corresponding time
+frame, which consists of a time point and a period length. The time
+point gives the start of the timeseries; the period length gives the
+length of the period between consecutive values in the timeseries.
+Each parameter value is assumed to be the **average value** for its
+corresponding period. This implies that additive values such as
```
znicholls:
```suggestion
corresponding period. This implies that values which describe the transport of a quantity such as
```

znicholls:
Only one comment, then if you can rebase and push I can merge this in

docs/usage.rst
swillner: I think this is too much about the interpretation of the values rather than their nature of not being averages...
```suggestion
corresponding period. This implies that values that are not averages but, for instance, absolute values, such as
```
?

znicholls: nice!

znicholls: although it's not clear to me why concentrations (ppm) are not absolute values but anyway let's go with this for now

swillner: hmmm, I will try to think about good names for the two types of values the next days...

znicholls: I swear state and flux variables is right but I must be making it up as google gives nothing...

swillner:
@znicholls Should be ready now ;)

znicholls:
wicked thanks

PR #111, opened 2019-02-10
Rewrite aggregation to collect list of children views

swillner:


openscm/parameter_views.py
```diff
@@ -199,32 +149,43 @@ def get_series(self) -> Sequence[float]:
...
-        return data
+    def _get_child_data_views(
+        self, parameter: _Parameter
+    ) -> Sequence["TimeseriesView"]:
+        if parameter._children:
+            return sum(
+                (self._get_child_data_views(p) for p in parameter._children.values()),
+                [],
+            )
+        return [
```
znicholls: if there are no children, but we ask for children, shouldn't we throw an error rather than just return self? Or is this needed for the recursion to work?

openscm/parameter_views.py
```diff
@@ -128,21 +68,26 @@ def get(self) -> float:
...
-            data += v.get()
-
-        return data
+    def _get_child_data_views(self, parameter: _Parameter) -> Sequence["ScalarView"]:
+        if parameter._children:
+            return sum(
+                (self._get_child_data_views(p) for p in parameter._children.values()),
+                [],
+            )
+        return [ScalarView(parameter, self._unit_converter._target)]
```
znicholls: if there are no children, but we ask for children, shouldn't we throw an error rather than just return self? Or is this needed for the recursion to work?

openscm/parameter_views.py
swillner: that's for the recursion; for the leaves of the parameter tree

znicholls: I find these methods which are named like `get_child_parameter` but then sometimes return self really confusing, is there another pattern/name we can use. Even just calling it `get_child_parameter_or_self`?

swillner: renamed it to ``_get_data_views_for_children_or_parameter`` and made it local as we do not need it as a class method

PR #112, opened 2019-02-12
Add FAIR adapter

znicholls:
Adds a first draft of an adapter for FaIR

- [x] Implementation finished
- [x] Tests added
- [ ] Documentation added


znicholls:
@lewisjared can you hit merge when you're happy with this (and if you cbf looking, just hit merge haha)

openscm/__init__.py
```diff
@@ -8,6 +8,7 @@
     convert_scmdataframe_to_core,
     convert_core_to_scmdataframe,
     convert_config_dict_to_parameter_set,
+    df_append,
 )
 from .adapters import get_adapter
 from .constants import ONE_YEAR_IN_S_INTEGER
```
swillner: @znicholls I thought you don't want to use fixed time intervals?

znicholls: I don't, this is a hack to see if things run (hence the PR into scmcallib). They run, but using the fixed time interval means you move off FaIR's native annual grid onto an evenly spaced one which seems to be having subtle effects on the results annoyingly. I think the solution is to make our timeseries stuff able to handle non-evenly spaced grids too.

lewisjared:
Do you still want to merge this or are you going to try 360-day calendars?

znicholls:
> Do you still want to merge this or are you going to try 360-day calendars?

Let's merge and I'll leave that for future

Issue #105, opened 2019-02-08
OpenSCM pythonic interface


swillner:
As discussed with @rgieseke a more pythonic view to the `Core` object and in particular the `ParameterSet`, i.e. relying on attributes rather than on functions such as `get_writable_view` would be desirable.
@znicholls @lewisjared Any wishes from the high-level API as that should rely on the mid-level/low-level API for timeframe and unit conversion as well as abstraction layer for communication with adapters.

znicholls:
My feeling to date has been that once we get things setup, there will be little interaction with the low level API (everything will be with the high level and the models) and so I'm not too fussed (I think I've got my head around this views business). More pythonic API for the core will certainly make it easier for new developers.

znicholls:
I have a converter from scmdataframes to Core in the `scmcallib` branch so I'd rather not have to re-write things to accomodate another level (although writing again would be pretty simple compared to writing the first time).

swillner:
I think we should change the wording for these levels. Overall, I would say there should be the following ways (interfaces/access points/...) to use OpenSCM:

- The Core API (standardized to be language agnostic): ``Core`` class
- A pythonic API: ``OpenSCM`` class
- ``ScmDataFrame`` as a convenience data structure to use; especially for IO with the standard file formats
- Additional utilities such as ensemble running (``Ensemble`` class), plotting functions, ...

Then we could abandon "low-level"/"high-level" wording.
@znicholls If you think that makes some kind of sense I would open a PR with restructured documentation...

znicholls:
To double check:

- adapters would only interact with `Core`?
- what is the difference between OpenSCM and ScmDataFrame? Data frames are Pythonic already

Would the workflow still be, user --> run --> ScmDataFrame --> Core --> Adapter --> Model --> Adapter --> Core --> ScmDataFrame --> user --> plotting

swillner:
>     * adapters would only interact with `Core`?

Yes, ``Core`` handles adapters from the outside. Adapters only talk to the ``ParameterSet`` (as part of the Core API) they receive from ``Core``.

>     * what is the difference between OpenSCM and ScmDataFrame? Data frames are Pythonic already

``OpenSCM`` would do the running and setting of individual parameters (more conveniently/pythonic than ``ParameterSet``), ``ScmDataFrame`` would handle input/output data if the users prefers more of a table view and can be handed over to ``OpenSCM`` for running the model.

[...]

PR #118, opened 2019-02-13
Restructure architecture

swillner:
...as in #105

- [x] Tests added
- [x] Documentation added


rgieseke:
What's the CODEOWNERS thing, is it for notifications?

swillner:
>What's the CODEOWNERS thing, is it for notifications?

It's to automatically add people as reviewers when certain files are changed (see the link in the file for documentation).

swillner:
Done with the new structure (Pythonic interface: #105)

README.rst
```diff
@@ -7,16 +7,16 @@ OpenSCM
...
 The **Open Simple Climate Model framework** unifies access to several
 simple climate models (SCMs). It defines a standard interface for
 getting and setting model parameters, input and output data as well as
 for running the models. Additionally, OpenSCM provides a standardized
 file format for these parameters and scenarios including functions for
-reading and writing such files. Its high-level interface further adds
-convenience functions and easily enables stochastic ensemble runs,
-e.g. for model tuning.
+reading and writing such files. It further adds convenience functions
+and easily enables stochastic ensemble runs, e.g. for model tuning.
```
znicholls:
```suggestion
and easily enables ensemble runs, e.g. for scenario assessment or model tuning.
```

znicholls: Just as an fyi, given the work elsewhere, I think putting model calibration into OpenSCM would be a mistake (it's a separate application with its own pitfalls and tricks). OpenSCM should be able to run an ensemble of parameters generated from a calibration library and analyse the output, but it should not have a function like `model.tune_to_target(target)`.

docs/usage.rst
```diff
@@ -3,18 +3,18 @@ Usage
...
 
-The **high-level** interface adds several convinience functions on top
-of the low-level interface. It allows for reading and writing
-parameters from and to a standardized and other file formats,
-including whole scenario definitions. Also, it provides functions for
-running ensembles of model runs and doing stochastic analyses and
-model tuning.
+The **Pythonic** interface adds several convenience functions on top
+of the core interface. It allows for reading and writing parameters
+from and to a standardized and other file formats, including whole
```
znicholls:
```suggestion
from and to standardized and other file formats, including whole
```

docs/usage.rst
```diff
@@ -3,18 +3,18 @@ Usage
...
+scenario definitions. Also, it provides functions for running
```
znicholls:
```suggestion
scenario definitions. It also provides functions for running
```

docs/usage.rst
```diff
@@ -3,18 +3,18 @@ Usage
...
+ensembles of model runs and doing stochastic analyses and model
```
znicholls:
```suggestion
ensembles of model runs.
```

docs/usage.rst
```diff
@@ -3,18 +3,18 @@ Usage
...
+tuning.
```
znicholls:
```suggestion

```

docs/usage.rst
znicholls: The term stochastic is misleading and I think we should get rid of it

docs/usage.rst
```diff
@@ -40,10 +40,10 @@ parameter for emissions. Thus, it is identified by
 
     ``Emissions`` -> ``CO2`` -> ``Industrial``.
 
-In the low-level API parameters are expected to be identified by
-tuples of strings describing their position in the hierarchy, i.e. in
-this example ``("Emissions", "CO2", "Industrial")``. The high-level
-API also allows for giving it as a string with the hierarchical levels
+In the core API parameters are expected to be identified by tuples of
+strings describing their position in the hierarchy, i.e. in this
+example ``("Emissions", "CO2", "Industrial")``. The ScmDataFrame API
```
znicholls:
```suggestion
example ``("Emissions", "CO2", "Industrial")``. 
```

Delete the whole ScmDataFrame API sentence. We'll explain how to access data when we write the ScmDataFrame API docs

docs/usage.rst
```diff
@@ -98,12 +96,11 @@ specifying the underlying SCM and start and end time:
 Setting input parameters
 ************************
 
-In the low-level API parameters are get and set through of subclasses
-of :class:`ParameterView`. While the values of the parameters are
-stored internaly, a :class:`ParameterView` provides an (always
-up-to-date) "view" of the corresponding parameter giving the parameter
-values in a specific unit and, in the case of timeseries, a specific
-time frame.
+In the core API parameters are get and set through of subclasses of
```
znicholls:
```suggestion
In the core API parameters are get and set through subclasses of
```

docs/usage.rst
```diff
@@ -98,12 +96,11 @@ specifying the underlying SCM and start and end time:
...
+:class:`ParameterView`. While the values of the parameters are stored
+internaly, a :class:`ParameterView` provides an (always up-to-date)
+"view" of the corresponding parameter giving the parameter values in a
```
znicholls:
```suggestion
"view" of the corresponding parameter and will always return the parameter values in a
```

openscm/scmdataframe.py
```diff
@@ -6,9 +6,13 @@
...
 class ScmDataFrameBase(IamDataFrame):
-    """This base is the class other libraries can subclass
+    """
+    This base is the class other libraries can subclass.
 
-    Having such a subclass avoids a potential circularity where e.g. openscm imports ScmDataFrame as well as Pymagicc, but Pymagicc wants to import ScmDataFrame and hence to try and import ScmDataFrame you have to import ScmDataFrame itself (hence the circularity).
+    Having such a subclass avoids a potential circularity where e.g. openscm imports
+    ScmDataFrame as well as Pymagicc, but Pymagicc wants to import ScmDataFrame and
+    hence to try and import ScmDataFrame you have to import ScmDataFrame itself (hence
+    the circularity).
```
znicholls: @swillner can you put this back in a separate file please? The comment above should explain why.

openscm/openscm.py
```diff
@@ -0,0 +1,11 @@
+from .core import Core
+
+
+class OpenSCM(Core):
+    """
+    OpenSCM class.
+
+    Represents model runs with a particular simple climate model.
```
znicholls: What do we actually want this class to do? Be the container for a single run? So it would have attribtues like `_adapter` (a, potentially already intialised, adapter instance), `_input` (a `ParameterSet` which gets passed to `_adapter._initialise_model_inputs` etc.) and `_output` (which reads from a `ParameterSet` which the adapter creates for output?). Then for ensembles, you just stitch lots of these together?

setup.py
```diff
@@ -71,18 +75,18 @@ def run_tests(self):
...
     version=versioneer.get_version(),
-    description=DESCRIPTION,
-    long_description=README_TEXT,
+    description="A unifying interface for Simple Climate Models",
+    long_description="\n".join(README_LINES),
     long_description_content_type="text/x-rst",
-    author=AUTHOR,
-    author_email=EMAIL,
-    url=URL,
+    author="Robert Gieseke, Jared Lewis, Zeb Nicholls, Sven Willner",
```
znicholls: Sorry @swillner I really don't understand why we've moved all these away from being defined in one place at the top of the file?

setup.py
```diff
@@ -1,45 +1,41 @@
...
-]
-
-requirements_extras = {
+REQUIREMENTS_EXTRAS = {
     "notebooks": REQUIREMENTS_NOTEBOOKS,
     "docs": REQUIREMENTS_DOCS,
     "tests": REQUIREMENTS_TESTS,
     "deploy": REQUIREMENTS_DEPLOY,
-    "dev": requirements_dev,
+    "dev": [
```
znicholls: Why not use the `requirements_dev` variable?

README.rst
swillner: a pity, but ok

openscm/scmdataframe.py
swillner: sounds very much like a problem in pymagicc that create the circular dependencies. but will split the file again.

openscm/openscm.py
swillner: that's the more pythonic convenience layer (though i would prefer to see the core interface as the standard way to use openscm). implementation will be a dedicated pull request.

setup.py
swillner: matter of taste like the issue above, i suppose. waiting for @rgieseke's decision ;)

setup.py
swillner: matter of taste, i prefer keeping definition and usage close if its used only once. @rgieseke you decide ;)

swillner: If you prefer the constants at the top of the file then use constants for each of these? Also for classifiers?

README.rst
znicholls: To clarify, we're still going to do it. It's just a separate library (putting it in one library makes it way too big). Hopefully that means it's not such a pity?

setup.py
znicholls: that would be my preference but ultimately it's no big deal so up to @rgieseke

openscm/scmdataframe.py
znicholls: > will split the file again.

Thank you

At the moment I think it's a problem for any adapter package that also wants to provide an OpenSCM interface directly. The reason is:

- OpenSCM is imported
- that imports `openscm.highlevel`
- which imports `openscm.adapters`
- which imports `<model-adapter-package>`, e.g. `pymagicc`
- which imports `openscm.highlevel` in order to get `ScmDataFrame`

and hence your circularity. Having said that, it might be solved thanks to the lazy imports of adapters because that means that you've already loaded `openscm.highlevel` when you load `<model-adapter-package>`. I'll try putting it back in one file when I add the MAGICC adapter and see what happens.

openscm/openscm.py
znicholls: Ok but it's the same as `Core`, just more pythonic? 

If yes, do we want to tweak the diagram in the docs so that it's clear that all that `OpenSCM` does is basically wrap `Core` in a more Pythonic way?

znicholls: Sorry I should probably look at the diagram properly before commenting. All good!

setup.py
swillner: As @rgieseke pointed me to the way pandas does it (https://github.com/pandas-dev/pandas/blob/master/setup.py#L200), I bow my head in light of the better argument ;)

rgieseke: I'd check what Numpy, Pandas do ...

https://github.com/pandas-dev/pandas/blob/master/setup.py#L200

https://github.com/numpy/numpy/blob/master/setup.py#L36

Numpy has a mix, but classifiers separately, I think it's useful to have everything settable in one place only. Makes it easier to use the rest as a template for the next project.

PR #121, opened 2019-02-16
Adapter interface suggestions

znicholls:
@swillner hopefully this shows what I'm thinking in terms of how to test the adapters.

I've moved the adapter tests from unit into integration as they'll have to call e.g. run and so will be testing how things integrate with the actual model calls and hence aren't unit tests. I've mocked out the adapter_registry tests to make them a bit more robust too. Lastly getting rid of the `"""` in the error message leaves the formatting up to the interpreter and doesn't prepend the error message with heaps of whitespace.


tests/integration/test_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+import pytest
+
+
+from openscm.adapter import Adapter
+from openscm.core import ParameterSet
+
+
+@pytest.fixture(scope="function")
+def test_adapter(request):
+    """Get an initialised instance of an the requesting classes ``tadapter`` property.
```
swillner: We are using different styles here. I prefer the line break after “”” as it looks much tidier and this whitespace won’t hurt in comments. Is there any standard?

znicholls:
> Can you adjust my old `AdapterTester` (i.e. delete it) and docs to your version here, please?

I just rebased so think that should be ok now?

tests/integration/test_adapter.py
znicholls: like so?

```python
"""
Get an initialised instance of an the requesting classes ``tadapter`` property.
"""
```

Happy for us to have that as the standard, I have no preference. black seems to be ambivalent to it so if we want to automate fixing it we might need to get fancier.

swillner:
I had added an `AdapterTester` in `tests/adapters/__init__.py` and recommended that folder for the adapter tests. Should we keep doing so?

swillner:
For the methods in the test class Codacy/PyLint recommends using class methods which seems reasonable here, these class are not instantiatd, are they?


```
    @classmethod
    def test_something(cls, test_adapter):
        ...
```

tests/integration/test_adapter.py
swillner: Cool, would like that ;)

znicholls:
> I had added an `AdapterTester` in `tests/adapters/__init__.py` and recommended that folder for the adapter tests. Should we keep doing so?

Ah ok yep I think so. So we'd:

- put `test_adapter` back into `unit` as that's literally just testing `openscm.Adapter`
- put all of our adapter tests in `tests/adapters`
- put a base adapter tester in `tests/adapters/base.py`
- put other adapter testers in their own file

One other subtle thing. As we're using `pytest`, none of our test folders should have `__init__.py` files as that blows up how `conftest.py` works. When there's a concrete example I'll explain better.

swillner:
> * put `test_adapter` back into `unit` as that's literally just testing `openscm.Adapter`
> * put all of our adapter tests in `tests/adapters`
> * put a base adapter tester in `tests/adapters/base.py`
> * put other adapter testers in their own file

Nice! Only, the third point makes the first redundant, isn’t it? Or what would you test in `base.py`?

> One other subtle thing. As we're using `pytest`, none of our test folders should have `__init__.py` files as that blows up how `conftest.py` works. When there's a concrete example I'll explain better.

Oh, good to know. I’m really a noob with pytest ;)

znicholls:
> Or what would you test in `base.py`?

Nothing, it's just where the base tester lives

swillner:
> > Or what would you test in `base.py`?
> 
> Nothing, it's just where the base tester lives

Ahhh, ok .

tests/integration/test_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+    outputstub = "Parameters"
+    Adapter.__abstractmethods__ = set()
+    adapter = Adapter(parametersstub, outputstub)  # pylint: disable=abstract-class-instantiated
+    assert adapter._parameters == parametersstub
+    assert adapter._output == outputstub
+
+
+def test_adapter_base_class_initialise_run_parameters():
+    Adapter.__abstractmethods__ = set()
+    adapter = Adapter(ParameterSet(), ParameterSet())  # noqa
```
swillner:
```suggestion
    adapter = Adapter(ParameterSet(), ParameterSet())  # pylint: disable=abstract-class-instantiated
```

tests/integration/test_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+    start_time = 0
+    stop_time = 1
+    adapter.initialize_run_parameters(start_time, stop_time)
+    assert adapter._start_time == start_time
+    assert adapter._stop_time == stop_time
+
+
+def test_adapter_base_class_run():
+    Adapter.__abstractmethods__ = set()
+    adapter = Adapter(ParameterSet(), ParameterSet())  # noqa
```
swillner:
```suggestion
    adapter = Adapter(ParameterSet(), ParameterSet())  # pylint: disable=abstract-class-instantiated
```

tests/integration/test_adapter.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Extra tests can be adapted depending on what the adapter should actually
+        do on shutdown.
+        """
+        test_adapter.shutdown()
+
+    def test_initialise_model_input(self, test_adapter, test_config_paraset):
+        """Test that initalising model input does as intended
+        """
+        # @swillner I am not sure how we want this to work. I still don't understand
```
swillner: I think that should be up to the model/adapter. Maybe we should also implement a way for the adapter to tell, e.g. for an ensemble run, if it would like to change the model input or the run parameters faster in the ensembke loop.

tests/unit/test_adapter_registry.py
```diff
@@ -1,11 +1,28 @@
...
+    assert load_adapter("stub") == 1
+
+
+def test_adapter_registry_unknown_model():
     with pytest.raises(KeyError, match="Unknown model 'unknown'"):
         load_adapter("unknown")
+
+
+def test_adapter_registry_import_error():
+    error_msg = re.escape(
```
swillner: Why `re.escape`?

tests/integration/test_adapter.py
swillner: damn, too slow with my review ;)

tests/integration/test_adapter.py
swillner: Sorry, the noqa didn’t work...

tests/unit/test_adapter_registry.py
znicholls: from memory, pytests's match needs to have things escaped if you special characters like square brackets and spaces and new lines

tests/integration/test_adapter.py
znicholls: > Maybe we should also implement a way for the adapter to tell

That sounds good, thoughts on how to do it that don't require passing a parameter set to these calls? A flag (which by default signals that it's a run parameter but the user can switch to model input for speed if they want?)

swillner: I think we need both, one for the adapter indicating the preference of order of these two functions (in an ensemble loop), and something indicating for the paramters. The latter we could specify as a standard as that should be model independent, isn’t it?

znicholls: > I think we need both, one for the adapter indicating the preference of order of these two functions (in an ensemble loop)

I think it will always be `set_model_inputs`, then `set_run_parameters` (timeseries are longer and involve more data than configuration flags so you want to do timeseries setting less often). 

> The latter we could specify as a standard as that should be model independent, isn’t it?

Yep!

So perhaps we have two standards to add. 1) That models know that `set_model_inputs` should be used for 'scenario/driver setting' and so would be called less often than `set_run_parameters` which would be used for 'config setting'. 2) Parameters have a flag indicating whether they should be set when `set_model_inputs` is called or whether they should be set when `set_run_parameters` is called and adapters have to know how to handle this?

swillner: cool. If we set a standard we don’t need a flag, it’s up to the user/adapter dev to adhere to them

swillner:
> For the methods in the test class Codacy/PyLint recommends using class methods which seems reasonable here, these class are not instantiatd, are they?
> 
> ```
>     @classmethod
>     def test_something(cls, test_adapter):
>         ...
> ```

@znicholls just pinging you on this

tests/adapters/test_magicc.py
```diff
@@ -0,0 +1,13 @@
+# Example file, would look something like this
```
swillner: Let’s have this in the docs and only include magicc tests with the magicc adapter later

tests/integration/test_adapter.py
znicholls: Ok I'll leave that to you to as I still don't really get how to implement it

znicholls:
> @znicholls just pinging you on this

Nice thans

swillner:
Awesome! Could you just remove the rcp *.DAT files that sneaked in in the last commit and I'll be happy to merge this!

znicholls:
Damn it will do in the morning

znicholls:
@swillner hopefully this is now gtg

swillner:
Hooray! ;)

Issue #51, opened 2018-11-13
Timeframe conversion: What about points w/o data?


swillner:
When converting to a timeframe for which there are gaps in the data (e.g. the target timeframe starts way before the source timeframe) we have at least three possibilities:
- Raising an error (currently implemented)
- Extrapolating the data (how would we want to do that?)
- Returning ``NaN`` for gaps in the data

PR #125, opened 2019-02-18
Timeseries update

znicholls:
This PR updates OpenSCM's handling of timeseries to cover a larger set of use cases.

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


znicholls:
@swillner can you have a look at `notebooks/user-timeframes.ipynb` and let me know what you think please?

swillner:
- Hmm, I thought we already decided for linear interpolation. But, of course, we could also provide quadratic/cubic/... interpolation.

- Extrapolation would only occur at the edges of the time frame. I don't think we need to support more extrapolation, or do we?

- Currently, we assume average "flux" values, so, yes, they should be integral preserving.

- The OpenSCM function does not work because it needs `int` values as time steps (seconds)

- Why make a new notebook for that?

znicholls:
>     * I thought we already decided for linear interpolation

We may have. If so, let's make it clear in the docs?

>     * I don't think we need to support more extrapolation, or do we?

No I don't think so either but there's nonetheless a choice about whether extrapolation is constant, or a linear extrapolation, or an extrapolation of the last segment of your spline etc.

>     * we assume average "flux" values

Yep, the point I am trying to make is that doesn't do sensible things in all cases. For example, an integral preserving interpolation of concentrations is almost never what the user will actually want.

>     * Why make a new notebook for that?

It was not clear to me how a user would ever be able to work with our timeseries/what wrangling they would have to do to get them following the right convention. This notebook will hopefully, eventually, be that but for now is just an easy place to explain.

swillner:
Ok, you are totally right, we were overly ridig so far.
- The main differences are between point data vs average data, not state variables vs flux variables. In principal, one can give concentrations as average data and emissions as point data. It is just not intuitive.
- For point data, of course, the interpolation is straight forward, be it linear/cubic/whatever, though also here are assumptions being made (e.g. that the step-wise linear functions change exactly at the given data points).
- In general, I suppose the problem is that dealing with average data is much less intuitive than handling point data. Saying that we assume "left-edge" points is missleading as the value for the time intervals is always for the whole interval, not at some point in it. That is also why plotting them as step functions is not exactly correct (though the canonical way to do so). When doing the interval preserving interpolation (which is a must for average data) the interpolated points are in general, of course, different than the original ones (i.e. when simply connecting them as with point data, they do not intersect at the original points).
- I agree with you that we should make the assumptions much clearer and they should be given in the docs rather than being hidden in some notebook.
- To allow for more flexibility, I think, we should not make it part of the standard in which form particular time series are given (point data/average data), but require the user to give that information when setting up the parameter views (just like giving the time frame information, which we have to advance for non-equidistant time points anyway). Conversion between point/average data can be done by the library when necessary with the option for the end-user to control the way of that conversion as well as the interpolation method used when changing time points. That would mean defining a `TimeFrame` would require to give information on:
  - Time points (either start+step length or explicit, possibly non-equidistant points)
  - Is the data point or average data (former implies not-integral preserving, latter implies integral preserving); for point data I would say these have to be at the exact time points giving (no left-edge/... needed)
  - Interpolation method: linear, cubic, others? (for each defining our assumptions in the docs. By its nature we cannot do interpolation without some assumptions).
  - Maybe giving information on extrapolation method for the edge cases. I would not support extrapolation beyond the edges.
- In my opinion, the user can and should not have to do the interpolation themselves. But as with knowing which parameters have been used by a model adapter, we should probably along the road also incorporate methods that give the data exactly as the model used it (just to be transparent).
- Another point that is still open with time series is how to deal with missing data (`NaN`s in between), see #51.

Does that make sense? ;)

znicholls:
>     * In principal, one can give concentrations as average data and emissions as point data. It is just not intuitive.

Ah yes, you're spot on there

>     * That is also why plotting them as step functions is not exactly correct (though the canonical way to do so).

Mmm true. But we do need to make some assumption about how they behave in the middle in order to do the interpolation. 

>     * Interpolation method: linear, cubic, others?

Yep I think linear and cubic would be a good pair

>     * for point data I would say these have to be at the exact time points giving (no left-edge/... needed)

Yep I like it

>     * Maybe giving information on extrapolation method for the edge cases.

Yep or potentially no extrapolation by default, but offer extension if user really wants?

>     * we should probably along the road also incorporate methods that give the data exactly as the model used it (just to be transparent).

Yep sounds good. 

>     * Another point that is still open with time series is how to deal with missing data (`NaN`s in between),

I think error or nan is best.

> Does that make sense? ;)

Yep. 

There's a lot to do haha What do you think of the following?

- double check our existing notebook (flip the order so we illustrate how our integral preserving interpolation currently works, I'm not actually sure it is integral preserving anymore...)
- discuss options for what other options we want to offer the user
- add options for interpolation of point data

As part of all of the above, we'll have to implement the timeframe information given above.

swillner:
> > * Another point that is still open with time series is how to deal with missing data (`NaN`s in between),
> > ```
> 
> I think error or nan is best.

I mean, how do we deal with `NaN`s when interpolating/extrapolating? Just "pipe it through" and let `NaN`s come out of the whole procedure?

>     * double check our existing notebook (flip the order so we illustrate how our integral preserving interpolation currently works, I'm not actually sure it is integral preserving anymore...)

Cool! Could you do that in the exisiting `timeframes.ipynb`?

>     * discuss options for what other options we want to offer the user
>
> As part of all of the above, we'll have to implement the timeframe information given above.

Will open a PR today and we'll continue discussion there?

znicholls:
> I mean, how do we deal with `NaN`s when interpolating/extrapolating? Just "pipe it through" and let `NaN`s come out of the whole procedure?

I would overwrite them. If you're interpolating, you probably want to get rid of the nans...

> Cool! Could you do that in the exisiting `timeframes.ipynb`?

Yep

> Will open a PR today and we'll continue discussion there?

Sounds good

swillner:
@znicholls Interesting. What's your rationale behind the renaming? The `Timeframe`/`Timeseries` class only describes the framing of a time series (i.e. its time points) but is not a time series itself.

znicholls:
> What's your rationale behind the renaming?

Timeframe means the time over which the data applies and to me implies more start time to end time, not that this class is for handling timeseries data/interpolating and extrapolating. I'm not sure `Timeseries` is right either but did it more just to see how hard such a global renaming is: answer is, not very :)

> `Timeframe`/`Timeseries` class only describes the framing of a time series

What does this mean exactly? It only handles the time points but not the data? I still feel like all we should need are views and interpolators?

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# OpenSCM Timeseries\n",
+    "\n",
+    "In OpenSCM, the class for handling timeseries is `openscm.timeseries.Timeseries`. \n",
+    "\n",
+    "A timeseries in OpenSCM is always initialised from arrays i.e. discrete data which are associated with specific points in time. In this notebook we call the initialisation values the 'source data' and the times with which they are associated the 'source times'.\n",
```
swillner: By definition a timeseries is always discrete...

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "\n",
+    "To estimate what the value of the variable under consideration is at points in time other than the source points (call these points the 'target times'), i.e. to interpolate or extrapolate, we always take the following steps:\n",
+    "\n",
+    "1. start with the data which was originally used to set the timeseries\n",
+    "1. convert the discrete data to continuous data using an interpolation/extrapolation technique\n",
```
swillner:
```suggestion
    "2. convert the discrete data to continuous data using an interpolation/extrapolation technique\n",
```

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "    - e.g. linear interpolation, cubic spline, moving average\n",
+    "1. use the continuous data to determine the values at the target times aka the 'target values'\n",
```
swillner:
```suggestion
    "3. use the continuous data to determine the values at the target times aka the 'target values'\n",
```

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "\n",
+    "There is one additional complication in this process. It is the different conventions that are used with timeseries. For example, sometimes the data value is the value only at the exact point in time for which it is supplied while other times the data value represents the average of the underlying continuous data set over a given time period. These conventions affect how the conversion between continuous and discrete data is done and hence are vital to document and understand.\n",
+    "\n",
+    "*Note:* In order to recover the original data from the target data, you need to know which assumptions have been made in the interpolation process and what convention is being used.\n",
```
swillner: Do we need this note?

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+    "\n",
+    "<div class=\"alert alert-warning\">\n",
+    "  <strong>Warning!</strong> Emissions data is ocassionally reported without the rate e.g. 2100 emissions were 3 GtC. In OpenSCM, we explicitly do not allow this as it doesn't actually make sense. The cumulative emissions for the year were 3 GtC but the emissions were not 3 GtC at any given point in 2100. Before passing emissions data into OpenSCM it should have the units converted to rates (of course, cumulative emissions data should not be reported as rates and so perhaps the only fix required is changing the variable name from `Emissions` to `Cumulative Emissions`).\n",
```
swillner:
```suggestion
    "  <strong>Warning!</strong> Emissions data is ocassionally reported without the rate e.g. 2100 emissions were 3 GtC. Before passing emissions data into OpenSCM it should have the units converted to rates (of course, cumulative emissions data should not be reported as rates and so perhaps the only fix required is changing the variable name from `Emissions` to `Cumulative Emissions`).\n",
```
...we just don't support it. We could, in principle, convert to rates, but setting a standard here is much easier.

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Conversion\n",
+    "\n",
+    "To convert from discrete to continuous data, we interpolate the data such that their average (i.e. integral over the period divided by the period length) equals the original source data. We use three points per period: at each period edge the value is assumed to equal the average of the values for the two adjacent periods and at the period mid point the value is adjusted such that the period average equals the source data value for that period. For the first and the last period the mid points are assumed to be at the period average and the outer edges are adjusted accordingly.\n",
+    "\n",
+    "In a picture it looks like the below. To make it easier to see how the interpolation is integral preserving, we plot the source data with lines which show the integral, following the assumption that the data values represent the average over the time period."
```
swillner:
```suggestion
    "In a picture it looks like the one below. To make it easier to see how the interpolation is integral preserving, we plot the source data with lines which show the integral, following the assumption that the data values represent the average over the time period."
```

notebooks/timeseries.ipynb
```diff
@@ -0,0 +1,568 @@
...
...Rqvyz7RFaLHuQBlzM0pY26u7TLUpq85Xr5MwNl5PWfPBrNihRMjRnTW2yzZBbu5pU80DMipqRrpdc7O0rIyy5c7cvz4l7RtW49vv91Pp05L2LnzksYvWEqIiODRgQNA+i8Bmux1jotLYvVqfzp2XMzmzWdxdGzLhQuzmD69l8YbzaC/75O8KCWLvudIH7oxkJUrB3D+6lM+cvLk7NlgbZelFn3fJ9kVlGXevD08e6YfQzRel32iK0TDuQApsbGkxJb8R90lTR9zpDwzwsFhLYGBIaxdOxgnp3bpj+thluwSIiK47+ODnJwMgJycrPEGZ6ZGjSzZunUUPj5fUL16RSZN+pWePX/g1KnbGpuBI9jNjbTUVEBzvwSkpqaxffsfdO78LUuWHKBDh/qcPDmD775zyHeO3ExeeHGRi5zmNNZY44WXytvV92MrO6VkUUoOR8e2HPAbjUVFY5yc1rF8+WG9nHXDCy8axlpjFmtY5PeXLsrv+Mo+RKNly13MIQkAACAASURBVDe1UFnRKOV9AvqRRTScC/Di9m1e3L6t7TLUpm85ku+bcP9/Dfj77//YtGkEn3zydtbP9C3Lq4Ld3JDTcv6HWRK9ztl17FifAwcmsW7dEGJjExk4cD1OTuuybjtdXJm/BJCSAqj/S4Asy5w4cQs7u+VMmbKdGjUq4ec3Hk/PkTRoUPjdpLzwwgUXEkmfVSOUUFxwUfk/d30/trJTShal5AConhrF1mV29O//LitWHMHJyZ2IiBfaLktlme8vw9v/Uee2cZHfX7oor+MrKiqOGTN20aSJJZMm6fYQjUxKep/oQxbd/vxBeK2kTzF2k7AZDZDjDdi1fQxt2yprpoRwf/+s3uZMcnIy4f7+UITxwUEE0ZWuqm9YAj4Bkw8lLLaac35lCj17BlOhXyTmMx5j/GaS6uvKEBsbQoJnBFa3008jYQ2TwUDCNKYN5Sysi7SuhKtlebKoFvHnKmBcNwFLj8fEfHSFmdJelddxkYtZjeZMccQxm9k441ykeoTXW5HfXyqoSSKUgUc/lqF6x6pc+CqNd+1mYbkmBLNOMRrdVkn4//dXmazHlPj+mjdvN0+fxrBly+c6P0RD0A5xVAg64cqVUBYu/I2LF+9iXC+Nmt53adtYWY1mALsLFwCIDAwEind3pIEMLPb2JROZKp8/oaLDM56vqc7zDW8Qs78yl
```
swillner: I think we should make clear above that this is only to show how the interpolation is done and that protected members should not be used by the end-user.

swillner:
```suggestion
    "This result may seem non-intuitive at a first glance, but is due to the preservation of the integrals and, thus, different to a more intuitive linear interpolation."
```

openscm/timeseries.py
```diff
@@ -150,34 +157,38 @@ def _calc_linearization_values(values: np.ndarray) -> np.ndarray:
     )
 
 
-def _calc_linearization(
-    values: np.ndarray, timeframe: Timeframe
+def _calc_continuous(
```
swillner: We should rename this function (though it's going to change in #126 anyway). `_calc_stepwise_linear_function`?

tests/unit/test_timeframes.py
```diff
@@ -36,20 +36,20 @@
         7.25,
         10.75,
     ],
-    (3, 3, 0, 5, 0): [-0.77777778, 4.13333333, 4.13333333, 5.51666667, 5.05952381],
+    (3, 3, 0, 5, 0): [-1.66666667, 4.13333333, 4.13333333, 5.51666667, 3.01666667],
```
swillner: Why do these differences at the edges occur?

swillner:
> > What's your rationale behind the renaming?
> 
> Timeframe means the time over which the data applies and to me implies more start time to end time, not that this class is for handling timeseries data/interpolating and extrapolating. I'm not sure `Timeseries` is right either but did it more just to see how hard such a global renaming is: answer is, not very :)
> 
> > `Timeframe`/`Timeseries` class only describes the framing of a time series
> 
> What does this mean exactly? It only handles the time points but not the data? I still feel like all we should need are views and interpolators?

In my eyes, time "frame"/"framing" means more than just start and end point. But you are right, we should rename that. Right now it describes the time points of a time series (as opposed to the data values), but also handles the interpolation internally. We could, of course, incorporate the time points information in the view and then have the description of the interpolation (and the internal actual conversion) separate. Does that feel more naturally to you?

znicholls:
> Does that feel more naturally to you?

Yep that would be my preference. Basically keep data and time points together, then pass them somewhere else for interpolation and extrapolation.

tests/unit/test_timeframes.py
znicholls: Not sure. I did the calculations by hand and am pretty sure that these numbers are correct for linear extrapolation. Perhaps with constant extrapolation you would get the other numbers? I thought we'd tried to implement linear extrapolation last time so maybe there was simply a bug...

openscm/timeseries.py
znicholls: The way it's implemented now it's not calculating a stepwise linear function, it's calculating a continuous linear interpolation. I think a name which makes it clear that we're calculating a continuous function here would be useful (and then maybe we add args for how to make things continuous that can take arguments like 'linear' or 'cubic' and 'integral_preserving' or not).

notebooks/timeseries.ipynb
znicholls: Ok how about a note at the top like, "This notebook accesses a number of private methods and attributes of OpenSCM classes. We do this to illustrate how OpenSCM works under the hood. However, the user should not access these private classes as these APIs are likely to change without warning. If you think one of these private classes should become public, please raise an issue at https://github.com/openclimatedata/openscm/issues."

notebooks/timeseries.ipynb
znicholls: I don't know. I found it helpful as I always it should be easy to go back and forth but have only just realised it's not. Maybe it's obvious to everyone except me, in which case I'm happy to remove it.

notebooks/timeseries.ipynb
znicholls: This should render as a 3 automatically?

notebooks/timeseries.ipynb
znicholls: This should render as a 2 automatically?

notebooks/timeseries.ipynb
znicholls: oops

znicholls:
> Nice, looks much simpler with `scipy` and I guess that's a dependency that we will have at some point, anyway ;)

haha yes I think so. The one problem with scipy's integration stuff is that it's numerical and hence slow. Hopefully the caching deals with that problem for us.

> But we should definitely find a better new name for timeframes.

Yep. Maybe when we split it up it will make more sense. The time points will go as an attribute of a `TimeseriesView` and then we just have an `Interpolator` or `TimeseriesConverter` class and hence can drop the `Timeframes` class altogether.

swillner:
> haha yes I think so. The one problem with scipy's integration stuff is that it's numerical and hence slow. Hopefully the caching deals with that problem for us.

Oh, that's a strong argument against this way as it's fairly easy to do the analytical integration here. Can we find a middle ground by only replacing the numerical integration part?

> Yep. Maybe when we split it up it will make more sense. The time points will go as an attribute of a `TimeseriesView` and then we just have an `Interpolator` or `TimeseriesConverter` class and hence can drop the `Timeframes` class altogether.

Ok, let's do so. Here? In #126?

notebooks/timeseries.ipynb
swillner: ah, cool. sorry, didn't know that ;)

notebooks/timeseries.ipynb
swillner: I would remove it, but have no strong feelings about that ;)

notebooks/timeseries.ipynb
swillner: Perfect!

tests/unit/test_timeframes.py
swillner: Ok, so if these follow the edge point extrapolation as described in the notebook, everything's fine.

notebooks/timeseries.ipynb
znicholls: @rgieseke thoughts? Do we keep this note about why going back and forth isn't that trivial or delete?

znicholls:
> Can we find a middle ground by only replacing the numerical integration part?

Yes but then the only interpolation we can offer that is mean preserving is linear (if you use the trapezium rule you're implicitly using linear interpolation between your points). I'll change it and update the notebook to make it as clear as possible.

> Regarding the renaming issue remaining, I think we can keep these new names and do the restructuring in another PR.

Ok cool

znicholls:
@swillner would it be possible for you to make a PR into here which puts the caching back? I'm not sure how it was tested but I could delete the entire `_Interpolation` class without failing any tests so think we need to think that one through again. What is the point of the caching i.e. what is the expensive step?

swillner:
> > Can we find a middle ground by only replacing the numerical integration part?
> 
> Yes but then the only interpolation we can offer that is mean preserving is linear (if you use the trapezium rule you're implicitly using linear interpolation between your points). If we want to add anymore options they need to be custom (as analytic solutions always are) and so we lose a lot of flexibility as a result.

I mean, we never mix interpolations, so we could do analytical interpolation depending on which kind of interpolation we do, so no loss of flexibility here. Nevertheless, maybe the loss of performance is not that bad here, it's just a pity of my efforts to keep exactly that ;)

> @swillner would it be possible for you to make a PR into here which puts the caching back? I'm not sure how it was tested but I could delete the entire `_Interpolation` class without failing any tests so think we need to think that one through again. What is the point of the caching i.e. what is the expensive step?

What caching do you mean? We could and should easily add caching for getting time series data depending on if it has changed or not. Do you mean that?
Caching views is primarily for the lookup in the parameter/region hierarchy.

znicholls:
> I mean, we never mix interpolations, so we could do analytical interpolation depending on which kind of interpolation we do, so no loss of flexibility here

No, what I mean is that if you have a cubic interpolation between 'target points' (for want of a better phrase), then the trapezium rule is only an approximation of the integral. To get the exact solution, you'd have to write something that works for the cubic interpolation. The same would be if you added any other type of interpolation in future. Using scipy allows you to just throw a continuous function (i.e. interpolation) at `quad` and it works out the rest. 

> What caching do you mean?

I mean the caching when doing conversion. Previously, the `_Interpolation` class used to cache the interpolation points but not the actual calculation of target values. My questions are:

1. Why did we cache that step (but not the actual calculation of target values)?
1. Depending on the answer to 1), what do we want to cache?
1. Can we put that back in as I've currently destroyed all caching?

swillner:
Regarding integration: Yes, every polynomial interpolation would be fairly straight forward. Will think about a good architecture for that. We will probably never have a non-polynomial interpolation.

Regarding caching: Ok, we should cache at least the final values, but also caching the interpolation points makes sense as these only depend on the two underlying (source and target) time points, not the values.

znicholls:
> every polynomial interpolation would be fairly straight forward

Perhaps but we still have to write and test them all ourselves (rather than using a pre-built, pre-tested library). This seems like a lot of work to avoid numerical integration no?

> Regarding caching: Ok, we should cache at least the final values, but also caching the interpolation points makes sense as these only depend on the two underlying (source and target) time points, not the values.

My question is why? Are these steps really so obviously expensive that's it's worth adding this extra complication before we know that it's actually worth it.

swillner:
ok, let's skip both for now and add them as TODO comments

swillner:
Maybe I'm too much about performance in a rather slow script language like python...

znicholls:
> Maybe I'm too much about performance in a rather slow script language like python...

I think you're right to worry. But let's get the thing working, with the tests we want first, then optimise. Doing it the other way round quickly makes life impossible (or at least that's my experience) and you can normally find 10x improvements with 2hrs and a decent profiler.

Should be good to go @swillner

PR #127, opened 2019-02-22
Fix adapter interface description and tests

swillner:


tests/unit/test_adapter_registry.py
```diff
@@ -21,12 +20,3 @@ def test_adapter_registry_unknown_model():
 
     with pytest.raises(KeyError, match="Unknown model 'unknown'"):
         load_adapter("unknown")
-
-
-def test_adapter_registry_import_error():
```
swillner: We will never have the possibility to test this as we include all dependencies already when doing `pip install .[tests]`. However, do we actually need to?

openscm/adapters/__init__.py
```diff
@@ -33,13 +33,14 @@ def load_adapter(name: str) -> type:
             if name == "MODELNAME":  # pragma: no cover
                 from .modelname import MODELNAME
 
-                adapter = MODELNAME(parameters)
+                adapter = MODELNAME
```
swillner: Also added a sentence to the docs clarifying that users should return their adapter class rather than an instance of that.

swillner:
@znicholls Very sorry, my bad: `pytest` seems not to use the `@classmethod` functions, so I turned them back to methods.

tests/unit/test_adapter_registry.py
znicholls: @swillner that's what the mocking is for, you can mock things so that they do whatever you want (in this case raise `ImportError` even though the module is actually installed). Can you put the test back in please?

swillner: Done

openscm/core.py
```diff
@@ -183,7 +183,7 @@ def get_timeseries_view(
 
         The parameter is created as a timeseries if not viewed so far.
         The length of the returned ParameterView's timeseries is adjusted such
-        that its last value corresponds to a time not exceeding the ``end_time``
+        that its last value corresponds to a time not exceeding the ``stop_time``
```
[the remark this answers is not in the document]
swillner: Done, thanks

openscm/adapters/__init__.py
```diff
@@ -52,5 +54,6 @@ def load_adapter(name: str) -> type:
         if adapter is None:
             raise KeyError("Unknown model '{}'".format(name))
         else:  # pragma: no cover
```
znicholls: this should already be hit by `tests/unit/test_adapter_registry` no?

swillner: No, we don't have a working model adapter yet (I added the TODO comment below)

znicholls:
nice @swillner !

PR #131, opened 2019-03-04
Re-add build and coverage badges, closes #130

### LINKED ISSUE #130: Badges
Can we get code coverage and build state badges back? They've always been a big source of confidence for me.

swillner:


znicholls:
Ok so we just hard code passing and 100% as we know that's the case?

swillner:
Yes, otherwise we could only have a dynamic badge with red/yellow/green for the github actions. But we ensure passing build and coverage on the master branch, anyway. Once we have a release at least the build badge will be redundant, probably.

znicholls:
Hmm ok how about we scrap the build/coverage badges and just add a badge which links to the actions page?

swillner:
Would be from here: https://github.com/CultureHQ/github-actions-badge
ok?

znicholls:
> ok?

Perfect :)

PR #133, opened 2019-03-06
Timeseries converter

swillner:
- [x] Implementation finished
- [x] Existing tests adjusted
- [x] Missing tests added
- [x] Documentation adjusted
- [x] Notebooks adjusted


openscm/core.py
```diff
@@ -256,19 +275,27 @@ def get_writable_timeseries_view(
...
         """
         Get a parameter or ``None`` if not found.
 
         Parameters
         ----------
         name
             :ref:`Hierarchical name <parameter-hierarchy>` of the parameter
-        region
+        region_name
             Hierarchical name of the region or ``()`` for "World".
```
znicholls:
```suggestion
            Hierarchical name of the region".
```

I thought we got rid of this?

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,364 @@
+"""
+Different climate models often use different time frames for their input and output
+data. This includes different 'meanings' of time steps (e.g. beginning vs middle of year) and
+different lengths of the time steps (e.g. years vs months). Accordingly, OpenSCM
+supports the conversion of timeseries data between such timeseriess, which is handled in
```
znicholls:
```suggestion
supports the conversion of timeseries data between such timeseries, which is handled in
```

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,364 @@
...
+this module. A thorough explaination of the procedure used is given in a dedicated
+`Jupyter Notebook
+<https://github.com/openclimatedata/openscm/blob/master/notebooks/timeseriess.ipynb>`_.
```
znicholls:
```suggestion
<https://github.com/openclimatedata/openscm/blob/master/notebooks/timeseries.ipynb>`_.
```

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,364 @@
...
+    int_averages = [np.nan] * len(target_intervals[:-1])
+    for i, l in enumerate(target_intervals[:-1]):
+        u = target_intervals[i + 1]
+        y, _ = integrate.quad(continuous_representation, l, u)
+        int_averages[i] = y / (u - l)
+
+    return np.array(int_averages)
+
+
+def _calc_linearization_values(values: np.ndarray) -> np.ndarray:
```
znicholls: I think this can be deleted no? It doesn't get used anywhere and is a duplicate of calculate continuous?

znicholls: I lie ignore that comment. Perhaps we can rename somehow? `_calc_integral_preserving_linear_interpolation`?

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,364 @@
...
+        extrapolation_type
+            Extrapolation type
+        """
+        self._source = np.array(source_time_points, copy=True)
+        self._target = np.array(target_time_points, copy=True)
+        self._timeseries_type = timeseries_type
+        self._interpolation_type = interpolation_type
+        self._extrapolation_type = extrapolation_type
+
+        if source_time_points[0] > target_time_points[1]:
```
znicholls: this will become conditional on extrapolation_type in future right?

.gitignore
```diff
@@ -17,3 +17,5 @@ dist/
 docs/_build
 venv/
 htmlcov/
+.pyre
+.pyre_configuration
```
znicholls: what is pyre?

openscm/timeseries_converter.py
swillner: Yes, will add a comment

.gitignore
swillner: A static type checker (http://pyre-check.org/), but I switched to Mypy (https://mypy.readthedocs.io/) now in #134 as it seems to be more complete and easier to configure

openscm/core.py
swillner: We missed this probably

openscm/core.py
```diff
@@ -214,17 +221,25 @@ def get_timeseries_view(
...
 
     def get_writable_timeseries_view(
         self,
         name: Tuple[str],
         region: Tuple[str],
         unit: str,
-        start_time: int,
-        period_length: int,
+        time_points: Tuple[int],
+        timeseries_type: ParameterType = ParameterType.AVERAGE_TIMESERIES,
```
znicholls: we should probably make the default point timeseries in future right? More intuitive?

swillner: yes. would also remove a default here for now if that's ok?

znicholls: good for me

swillner:
@znicholls Cool. Can you merge this and then #134 please?

.pylintrc
```diff
@@ -10,3 +10,9 @@ disable=
     too-many-locals,
     # handled by black
     format
+
+[SIMILARITIES]
+min-similarity-lines=6
```
znicholls: why is this so high?

CHANGELOG.rst
```diff
@@ -15,3 +15,4 @@ master
 - (`#44 <https://github.com/openclimatedata/openscm/pull/44>`_) Add timeframes module
 - (`#40 <https://github.com/openclimatedata/openscm/pull/40>`_) Add parameter handling in core module
 - (`#104 <https://github.com/openclimatedata/openscm/pull/104>`_) Define Adapter interface
+- (`#133 <https://github.com/openclimatedata/openscm/pull/104>`_) Cleanup and advance timeseries converter
```
znicholls: @swillner can you add a CHANGELOG entry for #134 please?

.pylintrc
swillner: The default is 4, I think. Reverted back to the default and added comments to avoid the false positives.

CHANGELOG.rst
swillner: done

.pylintrc
```diff
@@ -10,3 +10,8 @@ disable=
     too-many-locals,
     # handled by black
     format
+
+[SIMILARITIES]
+ignore-comments=no
+ignore-docstrings=yes
+ignore-imports=yes
```
znicholls: doesn't duplicate docstrings and imports suggest something could be done better?

openscm/core.py
```diff
@@ -256,20 +275,28 @@ def get_writable_timeseries_view(
...
-    def get_parameter_info(self, name: Tuple[str], region: Tuple[str]) -> ParameterInfo:
+        parameter.attempt_write(unit, timeseries_type, time_points)
+        return WritableTimeseriesView(
+            parameter,
+            unit,
+            time_points,
+            timeseries_type,
+            interpolation_type,
+            extrapolation_type,
+        )  # WritableTimeseriesView
```
znicholls:
```suggestion
        )
```

openscm/parameter_views.py
```diff
@@ -166,14 +185,21 @@ def get_data_views_for_children_or_parameter(
...
+        self._unit_converter = UnitConverter(cast(str, parameter._info._unit), unit)
         self._timeseries_converter = TimeseriesConverter(
-            parameter._info._timeseries, timeseries
-        )
+            parameter._info._time_points,
+            time_points,
+            timeseries_type,
+            interpolation_type,
+            extrapolation_type,
+        )  # TimeseriesConverter
```
znicholls:
```suggestion
        )
```

openscm/regions.py
```diff
@@ -2,7 +2,7 @@
 Handling of region information.
 """
 
-from typing import Dict, Tuple
+from typing import cast, Dict, Optional, Tuple
 from .errors import RegionAggregatedError
 from . import parameters  # pylint: disable=cyclic-import
```
znicholls: why isn't the cyclic-import a problem?

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        if len(values) < 3:
+            raise InsufficientDataError
+
+        if self._timeseries_type == ParameterType.AVERAGE_TIMESERIES:
+            return _calc_interval_averages(
+                self._calc_continuous_representation(source_time_points, values),
+                target_time_points,
+            )
+        # else
```
znicholls:
```suggestion

```

openscm/parameters.py
```diff
@@ -4,9 +4,8 @@
 
 # pylint: disable=unused-import,protected-access
```
znicholls: why do we have unused imports?

openscm/units.py
```diff
@@ -87,6 +87,10 @@
 """
 # pylint: disable=unused-import
```
znicholls: why do we have unused imports?

openscm/utils.py
```diff
@@ -5,8 +5,10 @@
 from typing import Tuple, Union
 import warnings
 
+_ensure_input_is_tuple_calls: int = 0
```
znicholls: why this change?

.pylintrc
swillner: In principle, yes, but especially imports can lead to false positives. Will change the docstrings, though.

openscm/core.py
swillner: Unfortunately, pylint only checks for duplicate lines, not statements. These comments are needed to avoid false positives of duplicate code (see last commit).

openscm/regions.py
swillner: Cyclic imports are not necessarily a problem in Python (because of the `sys.modules` caching). This import, however, is necessary for Sphinx to find the correct types. Would be happy to do that in a better way, but have not found any.

openscm/parameters.py
swillner: Sphinx (see comment above)

openscm/units.py
swillner: Sphinx (see comment above)

openscm/utils.py
swillner: Pylint and Mypy were complaining. Both ways are equally good, I suppose.

swillner:
In general, I think our Pylint configuration is still quite strict. I would prefer to keep it that way and only locally disable checks were absolutely necessary.

openscm/timeseries_converter.py
```diff
@@ -0,0 +1,355 @@
...
+                )
+                res = interpolate.interp1d(
+                    linearization_points,
+                    linearization_values,
+                    fill_value="extrapolate"
+                    if self._extrapolation_type == ExtrapolationType.LINEAR
+                    else None,
+                )  # type: Callable[[float], float]
+                return res
+        # else
```
swillner:
```suggestion
```

.pylintrc
swillner: The default of 4 lines unfortunately leads to a lot of false postives for duplicate docstrings (especially because of the Sphinx-lines such as `Raises` etc.)

znicholls: cool all good then

openscm/regions.py
znicholls: ah ok fair

openscm/utils.py
znicholls: ok. I have a particular hate for global variables which arose in another project so would prefer to avoid the global if possible. However if this is the only one I'll survive.

swillner: Totally agree with you on globals ;) I have found a way to avoid that, just a second.

swillner: Done

swillner:
@znicholls Just these small changes ;) Would you merge this, please?

znicholls:
wicked cheers @swillner

PR #134, opened 2019-03-06
Fix type annotations and enforce them in Github Actions

swillner:
Use [Mypy](https://mypy.readthedocs.io/) to enforce typing.


znicholls:
@swillner is it possible to split the typing stuff from the timeseries stuff please? Do the type annotations only apply if they're there, or will this force to provide type annotations for everything?

swillner:
Will wait with this till after the merge of #133. Currently, this enforces both correctness and usage of type annotations (ignoring external non-type libraries)

swillner:
@znicholls Ok, I target the PR#133 branch now

znicholls:
> usage of type annotations

This feels so wrong in Python, I'm not sure if I am ok with it. @rgieseke @lewisjared ?

swillner:
>> enforces... usage of type annotations ...

Still it's just STATIC checking and only annotation, so making the code more robust. We can still use arbitrary duck typing using `Any` (see Mypy docs), though in most cases specifying the type requires almost no additional efforts.

swillner:
https://stackoverflow.com/questions/42460289/why-to-use-types-in-python-3-5

rgieseke:
Might increase barriers to contribute ... Could this be handled like code coverage? Something where 100% is nice, but might not always be possible?

swillner:
Hmm, giving a share of "missed" lines would be difficult in this case. We could nevertheless be less strict here. On the other hand I believe that these will be used incrisingly in the future (they are a fairly recent feature), even by scientists, I hope ;)

swillner:
We could have mypy in the makefile and encourage its use without enforcing it. Though, I'm afraid that would lead to many people just not caring about these things.

swillner:
What about a compromise enforcing type annotations to be correct, but only if they are given and only warn about missing type annotations without enforcing their completeness?

znicholls:
> What about a compromise enforcing type annotations to be correct, but only if they are given and only warn about missing type annotations without enforcing their completeness?

Sounds good to me.

swillner:
Ok, adjusted ;)

swillner:
@rgieseke @znicholls Ok, if I merge this now? It's going into the `timeseries-converter-advancement` branch anyway (#133).

swillner:
Wil wait with merging for #133 and then merge this into master.

rgieseke:
Sure, go ahead!

znicholls:
> Wil wait with merging for #133 and then merge this into master.

Ok cool. One thing though, could we make it practice to never merge our own PRs? Just good practice I think from what I've seen in other projects.

swillner:
@znicholls Merge this first, please

PR #135, opened 2019-03-06
Add clean-notebook makefile option

swillner:
As the metadata in the Jupyter notebooks (`*.ipynb`) is rather annoying when using git, now include a makefile target `clean-notebook` to clean the notebook metadata (to be done before committing notebook changes). It follows http://timstaley.co.uk/posts/making-git-and-jupyter-notebooks-play-nice/ and uses [jq](https://stedolan.github.io/jq/).


znicholls:
I think the implementation is very nice. However I really like having the output in the notebooks as then previews on github are actually useful. I'd be nervous about having notebooks with no output in master so I'd prefer to use this make target only on dev branches and to make it easier to see what's going on but would keep the output in the notebooks committed to master.

swillner:
Don't worry, it only strips the notebook of the metadata section (kernel name, etc.) and the execution counts. Output remains untouched (changed that from what is given on the linked website).

znicholls:
super nice

rgieseke:
Awesome! Could be something to have as part of the Actions workflow?

rgieseke:
This look really, really useful! Have you tried it with an opened notebook? What happens when you remove the metadata from an open notebook, will it just be overwritten on the next save?

swillner:
I have not tried it while it's open. I'm afraid Jupyter readds these data on save. Unfortunately, there seems not to be an option to skip saving of metadata (yet).
Regarding Actions: I am hesitant to modify the repository in the actions. But we could do so. Or just decline "unclean" notebooks.

swillner:
@rgieseke still need your review here as a code owner ;)

rgieseke:
> I have not tried it while it's open. I'm afraid Jupyter readds these data on save. Unfortunately, there seems not to be an option to skip saving of metadata (yet).

Another option could also be to run the Makefile task only on "clean" files (In the git sense).

Let' merge this and see how we use it :-)

PR #137, opened 2019-03-12
Split unit and integration tests and fix Makefile

znicholls:
Re-arrange location of timeseries converter unit and integration tests and fix Makefile

(All N/A)

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


swillner:
I don't think that split was necessary; the aspect of being `unit` tests probably depends on preferences here.

znicholls:
fair (https://martinfowler.com/bliki/UnitTest.html) I prefer to try and keep unit tests for tests which isolate a single function/method and put anything which requires calling other functions/methods into integration tests. It just means that only one unit test should fail when you change a given method, whilst many integration tests may fail, which can make keeping track of things a bit easier. The trade off is that the unit tests become super small so you don't necessarily use them on everything but that's ok.

PR #139, opened 2019-03-13
Define parameter hierarchy

### LINKED ISSUE #22: Define standard parameter hierarchy

znicholls:
Closes #22, all details are in that issue

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


swillner:
Just a few comments:

- We should definetely get rid of model-specific parameters here. Either they are really model-specific and thus not be in this list or - better - we find a proper way to map them to standard parameters in the corresponding adapter
- We should, because of conversion, not give concrete units here, but their _kinds_, e.g. `mass (carbon) / time`
- I would relax on the regions requirements (at least for now)
- Ok, if I also push to this branch along the way and we do a detailled review later?

znicholls:
>     * We should definetely get rid of model-specific parameters here

Do we have any here?

>     * We should, because of conversion, not give concrete units here, but their _kinds_

Oh that's what you meant. I figured if we gave e.g. 'GtC / yr' then people would know that 'MtCO2 / yr' was allowed thanks to OpenSCM's unit conversion. 

>     * I would relax on the regions requirements (at least for now)

They're not really requirements. It's more that if you give a model `World|R5ASIA` but it's global only then it should just e.g. throw a warning saying "I am a global only model so will aggregate all variables to the global level first and your regional detail won't be used."

> * Ok, if I also push to this branch along the way and we do a detailled review later?

I'd prefer to not push to the same branch if that's ok. But frequent PRs from our respective forks sounds good and we do detailed review later.

swillner:
> > ```
> > * We should definetely get rid of model-specific parameters here
> > ```
> 
> Do we have any here?
> 

All containing "MAGICC" ;)

> > * Ok, if I also push to this branch along the way and we do a detailled review later?
> 
> I'd prefer to not push to the same branch if that's ok. But frequent PRs from our respective forks sounds good and we do detailed review later.

ok

znicholls:
> All containing "MAGICC" ;)

That's true although they're sadly widely used

docs/parameters.rst
```diff
@@ -2,3 +2,276 @@
...
+-----------
+
+'Pre-industrial' refers to an unperturbed state of the climate.
+Individual adapters can translate this into whatever year they need to for their model, but they should do such translations with this definition in mind.
+
+'Reference period' refers to the period a given variable is reported relative to the mean of.
+For example, 'surface temperature relative to a 1961-1990 reference period' refers to surface temperatures relative to the mean of the period 1961-1990.
+Adapters can report variables without reference periods if these are not used internally by the models.
+However, if the model uses an internal reference period then this should be indicated in the reported variable names by appending ``(rel. to XXXX-YYYY)`` to the variable name.
+Hence, 'surface temperature relative to a 1961-1990 reference period' would become ``Surface Temperature (rel. to 1961-1990)`` [TODO: discuss with other authors how we want this to play out].
```
swillner: One of the more difficult tasks for us ;) Let's have a chat about this issue in April

docs/parameters.rst
```diff
@@ -2,3 +2,276 @@
...
+- ``Emissions|Halon1301`` (mass Halon1301 / time)
+- ``Emissions|Halon2402`` (mass Halon2402 / time)
+
+
+Concentrations
+**************
+
+Atmospheric concentrations should be of the form ``Atmospheric Concentrations|XXX`` e.g. ``Atmospheric Concentrations|CO2``.
+OpenSCM models should understand all of the following variables.
+Note that adding up all sub-categories of ``Atmospheric Concentrations`` will either fail as the units will not all be the same or will produce a basically nonsense number (but still be possible).
```
swillner: Do we need this comment? Maybe generalize it and move to the general "Aggregation" section (see my PR).

docs/parameters.rst
```diff
@@ -2,3 +2,276 @@
...
+
+
+Configuration
+-------------
+
+Each model will have its own set of configuration parameters and conventions.
+In OpenSCM we allow the user to pass these to and from the model via the adapter, following the model's own internal conventions for naming.
+However, we also insist that models understand the following configuration options.
+
+- ``ecs`` (K) - equilibrium climate sensitivity
```
swillner: Shouldn't we spell these out?

docs/parameters.rst
```diff
@@ -2,3 +2,276 @@
...
+- ``Land to Air Flux|CH4|Permafrost`` (mass methane / time)
+
+
+Other
+*****
+
+Other variables which should be recognised by OpenSCM adapters are given below.
+
+- ``Surface Temperature`` (K) - surface air temperature i.e. tas
+- ``Ocean Temperature`` (K) - surface ocean temperature i.e. tos
```
swillner: We could separate Surface and Ocean using the regions...

docs/parameters.rst
```diff
@@ -2,3 +2,276 @@
...
+- ``Radiative Forcing|Halon1211`` (W/m\*\*2)
+- ``Radiative Forcing|Halon1301`` (W/m\*\*2)
+- ``Radiative Forcing|Halon2402`` (W/m\*\*2)
+
+
+Material Fluxes
+***************
+
+These variables can be used to store the flux of material within the model.
+They should be of the form ``X to Y Flux`` where the material is flowing from ``X`` into ``Y`` (and hence negative values represent flows from ``Y`` into ``X``).
```
swillner: Maybe ``("Flux", "X", "Y")``?

docs/parameters.rst
znicholls: think this will be fixed in #150 no?

docs/parameters.rst
znicholls: how do you mean spell them out?

docs/parameters.rst
znicholls: I don't think that works because you still have 'tas' over ocean regions (we'd need a height dimension which we don't want to deal with...)

docs/parameters.rst
znicholls: Would be super confusing given the convention about variables being the sum of all variables one level below I think. I'm not convinced my solution is perfect either though so if we can think of something smarter that'd be ideal.

znicholls:
@swillner do we have any more to add to this list?

- [ ] decide how to handle rebased variables
- [ ] spelling out parameters like ecs (what do we mean by 'spell out')
- [ ] decide on convention for flux variables

swillner:
Sounds good!
I mean, should we name the ECS parameter "ECS" or "Equilibrium Climate Sensitivity"?

znicholls:
Ah, in full is a good idea until people get sick of it

docs/parameters.rst
znicholls: @swillner I would propose sticking with my solution, I don't have any better ideas.

docs/parameters.rst
znicholls: any thoughts on this @swillner. This convention was the best I could come up with. Maybe there are better, machine readable ways?

docs/parameters.rst
znicholls: we removed this statement anyway so all good

docs/parameters.rst
znicholls: make TODO reference a live issue

znicholls: now references #167

PR #140, opened 2019-03-13
Add basic metric conversions

znicholls:
Adds basic metric conversions to OpenSCM, allowing for easy conversion to/from CO2-equivalent emissions.

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


swillner:
Nice! Again, the CSV looks like a job for an in-code `dict` ;)

znicholls:
> Nice! Again, the CSV looks like a job for an in-code `dict` ;)

Yep definitely going to need a smarter solution. Let me get a minimal working example up then we can talk about how to do it sensibly.

znicholls:
@swillner and @lewisjared this is ready for your caching wizadry. The basic problem is that these tables are most easily edited as csvs/tables rather than python dictionaries so the question is, "Can we keep these in a format which is easily read by csv editors such as excel whilst also finding a way to not sacrifice speed?" My current guess would be to only load the contexts when we need them, not by default..?

swillner:
I really don't see the need for a tabular format, here, to be honest. It's just a couple of numbers and not necessarily all contexts include conversions for all gases, isn't it?
Can you also run `make clean-notebooks`, please?

znicholls:
> I really don't see the need for a tabular format, here, to be honest. It's just a couple of numbers and not necessarily all contexts include conversions for all gases, isn't it?

It's a couple of numbers in these examples, eventually we want to have a table like https://www.ghgprotocol.org/sites/default/files/ghgp/Global-Warming-Potential-Values%20%28Feb%2016%202016%29_1.pdf but with extra columns for other metrics so it'll be pretty big (~60 rows x ~10 columns). All metric contexts like these will include conversions for most gases.

> Can you also run `make clean-notebooks`, please?

oui oui

swillner:
Ah, ok. If you think CSV should be the way to go, then let's lazy-load it.

znicholls:
> If you think CSV should be the way to go, then let's lazy-load it.

I'm not sure what's best, I just don't know what other solutions there are for making this sort of data easily editable but also easy to use cause I've never really thought about it before

swillner:
hmm, this looks - at quick first glance - like it is lazy-loading the unit registry (which is needed for every unit conversion) rather than only the contexts (which might only be needed or used occasionally), right?

znicholls:
> right?

I believe so. My aim was to just shift things around enough that you can then add the detail to only load the contexts as needed. I think I am now there so can hand over to you to show me how to do this properly.

PR #141, opened 2019-03-13
Add checkers

swillner:
Add as checkers/formatters/linters:
- flake8
- isort
- pydocstyle

Also adjust GitHub Actions (including documentation build checks).
Adjust code accordingly.


swillner:
@znicholls I think this is good to go ;)

PR #142, opened 2019-03-13
Add boolean and string parameters

swillner:
- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


znicholls:
nice @swillner. I think we still have an unsolved problem though (maybe this needs to go in a new issue). We're going to need to need an extra, 'miscellaneous' (or similar) parameter type to deal with all the stuff that SCM developers use. For example, MAGICC has parameters which are vectors but not timeseries (a type not supported) and lists of strings (a type not supported), FaIR has matrices but not timeseries (a type we don't support) and I don't even know what other models use. We clearly need a way to support this, and the simplest solution is to have an 'Other' (or similar) parameter type as far as I can tell. Adding Boolean and String covers some use cases but I think we still need the 'Other' class and in the process of adding such a class it may be worth removing the Boolean and String classes again just for simplicity and to reduce the number of parameter types we have to support?

swillner:
Hmm, let's collect what we need here in a new issue first...

swillner:
Hmm, let's collect what we need here in a new issue first...

znicholls:
alright see #143

PR #145, opened 2019-03-14
Type-fix core and add generic parameter type

swillner:


znicholls:
@rgieseke @swillner over to you for rebasing, final review and merging

PR #148, opened 2019-03-20
Lazy load contexts

swillner:


notebooks/emissions-units-with-pint.ipynb
```diff
@@ -81,7 +81,7 @@
    "source": [
     "## Pint Unit Registry\n",
     "\n",
-    "The `unit_registry` which sits underneath all conversions can be accessed via `UnitConverter`'s `unit_registry` property."
+    "The `unit_registry` which sits underneath all conversions can be accessed via `UnitConverter`'s `unit_registry` property. Nevertheless, you should regularly have no need to access it directly."
```
znicholls:
```suggestion
    "The `unit_registry` which sits underneath all conversions can be accessed via `UnitConverter`'s `unit_registry` property. Nevertheless, you should have no need to access it directly."
```

openscm/units/__init__.py
```diff
@@ -1,9 +1,23 @@
 """
 Unit handling.
 
-Our unit handling library is `Pint <https://github.com/hgrecco/pint>`_. This allows us
-to define units simply as well as providing us with the ability to define contexts.
+Unit handling make use of the `Pint <https://github.com/hgrecco/pint>`_ library. This
```
znicholls:
```suggestion
Unit handling makes use of the `Pint <https://github.com/hgrecco/pint>`_ library. This
```

openscm/units/__init__.py
```diff
@@ -1,9 +1,23 @@
...
+allows to easily define units as well as using contexts of several conversion types.
```
znicholls:
```suggestion
allows us to easily define units as well as contexts. Contexts allow us to perform conversions which would not normally be allowed e.g. in the 'AR4GWP100' context we can convert from CO2 to CH4 using the AR4GWP100 equivalence metric.
```

openscm/units/__init__.py
```diff
@@ -1,9 +1,23 @@
...
 
+Do not use Pint with OpenSCM explicitely. Internally, units are used the following way:
```
znicholls:
```suggestion
In general, you should not use Pint with OpenSCM explicitly. As illustration of how units are used internally, we provide the following example:
```

openscm/units/__init__.py
```diff
@@ -89,40 +103,354 @@
...
+        self.define("degreeF = degF")
+        self.define("kt = 1000 * t")  # since kt is used for "knot" in the defaults
+
+        self.define("ppt = [concentrations]")
+        self.define("ppb = 1000 * ppt")
+        self.define("ppm = 1000 * ppb")
+
+    def enable_contexts(self, *names_or_contexts, **kwargs):
+        """
+        Overload pint's `enable_contexts` to load contexts once the first time some are
```
znicholls:
```suggestion
        Overload pint's `enable_contexts` to load contexts once (the first time they are
```

openscm/units/__init__.py
```diff
@@ -89,40 +103,354 @@
...
+        used to avoid (unnecessary) file operations on import.
```
znicholls:
```suggestion
        used) to avoid (unnecessary) file operations on import.
```

swillner:
@znicholls gerne ;) thanks for the quick review. just merge when it’s convenient for you, it’s your branch ;)

Issue #123, opened 2019-02-18
ParameterSet: Parameter versioning + Log used/unused parameters


swillner:

swillner:
- [ ] Parameter versioning
- [ ] Log used/unused parameters
- [ ] Way to retreive parameter values exactly as "seen" by the model adapter

znicholls:
When we do this, any run functions can have arguments like `unused_notification` which can specify what should happen if there are unused parameters e.g. do nothing, raise a warning, throw an error. 

A test would be something like

```python
# if you try to initialise and run with a timeseries that can't be used,
# you get a warning
assert raises_warning(run(
    junk_para=12,
    unused_notification="warn"
))
```

PR #149, opened 2019-03-21
Some thoughts on using adapters

znicholls:
@swillner this just gives you a few thoughts on how we might want to modify the adapter tests, in the process defining how we should interact with adapters a little better. I leave it here for your perusal and for our discussion and we can decide whether to delete it or flesh it out later.

### DIFF
--- a/tests/adapters/base.py
+++ b/tests/adapters/base.py
@@ -1,3 +1,5 @@
+from openscm.parameters import ParameterType
+
 class _AdapterTester:
     """
     Base class for adapter testing.
@@ -17,6 +19,44 @@ class _AdapterTester:
     def test_initialize(self, test_adapter):
         test_adapter._initialize_model()
 
+        # This comes from how test_adapter is initialised, we might want to get
+        # rid of this in future as it will interfere with defaults.
+        assert test_adapter._parameters.get_scalar_view(
+            ("rf2xco2",),
+            ("World",),
+            "W / m^2",
+        ).get() == 4
+
+        # Would it make sense for us to also pre-set the outputs with empty arrays?
+        # Then we could use these outputs to help us with separating model config
+        # from model timeseries. It may require a re-name, but that would also make
+        # it possible for the `initialize_model_input` and `initialize_run_parameters`
+        # methods to know which parameters they're meant to look at for the
+        # initialisation.
+        assert test_adapter._output.get_timeseries_view(
+            ("Emissions", "CO2"),
+            ("World",),
+            "GtCO2/yr",
+            [0, 1, 2],
+            ParameterType.AVERAGE_TIMESERIES,
+        ).is_empty
+
+        assert test_adapter._output.get_timeseries_view(
+            ("model specific para",),
+            ("model region",),
+            "model unit",
+            [0, 1],
+            ParameterType.POINT_TIMESERIES,
+        ).is_empty
+
+        # We could turn this into an abstract method so that implementers are forced
+        # to do an assertion like the below.
+        assert test_adapter._parameters.get_scalar_view(
+            ("model specific para",),
+            ("model region",),
+            "model unit",
+        ).get() == expected_default_value
+
     def test_shutdown(self, test_adapter):
         """
         Test the adapter can be shutdown.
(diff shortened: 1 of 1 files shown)


tests/adapters/base.py
```diff
@@ -17,6 +19,44 @@ class _AdapterTester:
...
+
+        assert test_adapter._output.get_timeseries_view(
+            ("model specific para",),
+            ("model region",),
+            "model unit",
+            [0, 1],
+            ParameterType.POINT_TIMESERIES,
+        ).is_empty
+
+        # We could turn this into an abstract method so that implementers are forced
```
swillner: yes, that should be left up to people implementing adapters

tests/adapters/base.py
```diff
@@ -30,55 +70,167 @@ def test_initialize_model_input(self, test_adapter):
...
+        ).set(set_values)
+
         assert not test_adapter._initialized
-        # TODO test for missing but mandatory parameter
-        test_adapter.initialize_model_input()
+        # start and end time should be set here as they're part of setting the
+        # timeseries views of emissions/concentrations/other drivers
+        test_adapter.initialize_model_input(start_time, end_time)
         assert test_adapter._initialized
+        # TODO test for start time and end time being passed properly
```
swillner: Hmm, don't know how we should test if they are used correctly

tests/adapters/base.py
```diff
@@ -30,55 +70,167 @@ def test_initialize_model_input(self, test_adapter):
...
+        assert test_adapter._output.get_writable_timeseries_view(
+            ("Emissions", "CO2"),
+            ("World",),
+            "GtC/yr",
+            [-10, 0, 10],
+            ParameterType.POINT_TIMESERIES,
+        ).get() == set_values
```
swillner: I don't think we need this test, it doesn't test anything about an adapter...

tests/adapters/base.py
```diff
@@ -30,55 +70,167 @@ def test_initialize_model_input(self, test_adapter):
...
+        test_adapter._output.get_writable_timeseries_view(
+            ("Emissions", "junk"),
+            ("World",),
+            "GtC/yr",
+            [-10, 0, 10],
+            ParameterType.POINT_TIMESERIES,
+        ).set(set_values)
+        # if you try to initialise and run with a timeseries that can't be used,
+        # you get a warning
+        assert raises_warning(test_adapter.initialize_model_input(
```
swillner: I'd very much prefer to have a method to get used/unused parameters rather than throwing warnings

tests/adapters/base.py
```diff
@@ -30,55 +70,167 @@ def test_initialize_model_input(self, test_adapter):
...
+            "ppm",
+            [298, 300, 310],
+            ParameterType.POINT_TIMESERIES,
+        ).set(set_values)
+        # If you initialise with a different run mode, the emissions get reset
+        # in order to avoid confusion
+        test_adapter.initialize_model_input(
+            start_time,
+            end_time,
+            run_mode="concentration driven"
```
swillner: Should definetely be a model-specific parameter rather than a parameter of the method (this will never be called directly by the user, anyway)

tests/adapters/base.py
znicholls: ye but I guess my question is should it be forced? i.e. if we make this abstract, then they have to add a test like the below, if we don't then they can pass all our tests without ever writing a test that is specific to their model..

tests/adapters/base.py
znicholls: I think this would have to be model specific (and so maybe this test method would have to be abstract) but e.g. with MAGICC, you'd test that the start and end times actually get passed into the cfg file which defines how the run is to be done

tests/adapters/base.py
znicholls: If we have the convention that adapters should put input timeseries straight into `_output` then it tests that the adapter actually does that. Maybe that's not what we want though?

tests/adapters/base.py
znicholls: I suspect this will end up being personal preference. I'm pretty relaxed, the reason I went for warning was that one of the most common complaints I've heard about these models is that they do things which aren't expected/don't match people's assumptions. If you throw a warning then people are forced to look at it, whereas if you have a method to get used/unused parameters then people will only work it out if they go looking. Given people could have gone looking already but didn't, it seems to me they need to have things raised in order for them to realise that things aren't working as they expected.

tests/adapters/base.py
znicholls: No, `run_mode` (or something similar) needs to be one of the key OpenSCM parameters and will be frequently used by users. One of the most annoying things about comparing SCMs at the moment is that they all have different default run modes. One of the most important jobs of the adapaters is to translate the `run_mode` into the right settings for the given model so it's trivial for users to run all models in 'emissions driven' or 'concentrations driven' or concentrations-emissions driven hybrid' mode at once so a like-with-like comparison can be made.

tests/adapters/base.py
znicholls: @swillner shall I make this abstract so the adapters have to implement a test like it?

tests/adapters/base.py
znicholls: let's have a play when we actually have some adapters to work with

tests/adapters/base.py
znicholls: Following other discussions, I agree so this warning can go elsewhere, see https://github.com/openclimatedata/openscm/issues/123#issuecomment-492926301

tests/adapters/base.py
znicholls: @swillner does the reason for having this argument here make sense now?

tests/adapters/base.py
znicholls: @swillner thoughts here?

znicholls:
served it's purpose for discussion, now closing

PR #150, opened 2019-03-22
Suggestions for parameter hierarchy

swillner:
- Use unit types throughout
- Use tables instead of lists
- Define gases once and reuse
- Adjust text


docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
 .. _standard-parameters:
 
-Parameters
-==========
+Standard parameters
+===================
 
-OpenSCM follows a number of conventions related to parameters, we document these
-In OpenSCM a 'parameter' is anything which is passed to a model e.g. |CO2| emissions, equilibrium climate sensitivity, aerosol forcing scaling.
+In OpenSCM a ':ref:`parameter <parameter-hierarchy>`' is any named input or output variable of a mode, e.g. |CO2| emissions, equilibrium climate sensitivity, aerosol forcing scaling.
```
znicholls:
```suggestion
In OpenSCM a ':ref:`parameter <parameter-hierarchy>`' is any named input or output variable of a model, e.g. |CO2| emissions, equilibrium climate sensitivity, aerosol forcing scaling.
```

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+As described :ref:`here <parameter-hierarchy>`, parameters are given in a hierarchy, e.g. ``Emissions`` -> ``CO2`` -> ``Industrial``.
 
 Simple climate models come in many different shapes and forms hence we do not expect them to all be able to do everything.
-However, to be included in OpenSCM they must be able to understand all of OpenSCM's key parameters where here, 'understand' means that their adapters should either be able to use them or should throw sensible warning/error messages if they are used.
+However, to be included in OpenSCM they should make sure their parameters fit into these standard parameters as far as possible to ensure models can be interchanged easily.
```
znicholls: I don't like this change. They have to understand all of the key parameters, where 'understand' means that their adapters should either be able to use them or should throw sensible warning/error messages if they are used. If they just blow up when given a key parameter then they're useless.

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+[TODO: discuss with other authors how we want this to play out]
 
 
-Variables
----------
+Aggregation
+-----------
 
-Variables in OpenSCM come as part of a hierarchy, separated by the ``|`` character.
+Parameters in OpenSCM come as part of a hierarchy, in the following separated by the ``|`` character.
```
znicholls:
```suggestion
Parameters in OpenSCM come as part of a hierarchy, separated by the ``|`` character.
```

They're always separated by this character no? Or do you envisage being able to separate different hierarchical separators?

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
-- ``Emissions|HCFC141b`` (mass HCFC141b / time)
-- ``Emissions|HCFC142b`` (mass HCFC142b / time)
-- ``Emissions|SF6`` (mass SF6 / time)
-- ``Emissions|SO2F2`` (mass SO2F2 / time)
-- ``Emissions|Halon1202`` (mass Halon1202 / time)
-- ``Emissions|Halon1211`` (mass Halon1211 / time)
-- ``Emissions|Halon1301`` (mass Halon1301 / time)
-- ``Emissions|Halon2402`` (mass Halon2402 / time)
+As far as it makes sense, parameters that are higher in the hierarchy (e.g. ``Emissions|CO2`` is 'higher' than ``Emissions|CO2|Energy``) are the sum of all the variables which are one level below them in the hierarchy.
+For example, for ``Emissions|CO2|Energy``, ``Emissions|CO2|Transport`` and ``Emissions|CO2|Agriculture`` given, ``Emissions|CO2`` would be the sum of these.
```
znicholls:
```suggestion
For example, if ``Emissions|CO2|Energy``, ``Emissions|CO2|Transport`` and ``Emissions|CO2|Agriculture`` are provided, ``Emissions|CO2`` would be the sum of these.
```

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
-- ``Radiative Forcing|SF6`` (W/m\*\*2)
-- ``Radiative Forcing|SO2F2`` (W/m\*\*2)
-- ``Radiative Forcing|Halon1202`` (W/m\*\*2)
-- ``Radiative Forcing|Halon1211`` (W/m\*\*2)
-- ``Radiative Forcing|Halon1301`` (W/m\*\*2)
-- ``Radiative Forcing|Halon2402`` (W/m\*\*2)
+Standard parameters
+*******************
+
+Below we provide a list of the OpenSCM standard parameters adapters must adhere to as far as a specific variable concerns them.
```
znicholls:
```suggestion
Below we provide a list of the OpenSCM standard parameters adapters must understand to as far as a specific variable concerns them.
```

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+Alongside we give the type of unit that the parameter should be given in and how it should be expected by adapters.
+Conversion between particular units is done automatically if possible.
+
+In the following, ``<GAS>`` can be one of the standard :ref:`gases`.
+
+.. csv-table:: Standard parameters
+    :header: "Parameter name 0", "Parameter name 1", "Parameter name 2", "Unit type", "Note"
+
+    ``Emissions``, ``<GAS>``,, "mass <GAS> / time"
+    ``Atmospheric Concentrations``, ``<GAS>``,, "concentration (e.g. parts per X)", "Aggregation possible, but does not always make sense"
```
znicholls:
```suggestion
    ``Atmospheric Concentrations``, ``<GAS>``,, "parts per X where X is million, billion, trillion etc.", "Aggregation possible, but does not always make sense"
```

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+    ``Pool``, ``<GAS>``,, "mass <GAS>"
```
znicholls: I wonder if we can do better than 'Pool', maybe not...

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+    ``Transient Climate Response``,,, "temperature"
+    ``f2xco2``,,, "power / area", "Radiative forcing due to a doubling of atmospheric |CO2| concentrations from pre-industrial level"
+
+.. _gases:
+
+Gases
+*****
 
+.. csv-table:: Gases
+    :header: "Name", "Description"
```
znicholls: do we want an 'also known as' column here? For example 'HFC365mfc' is the same as 'HFC365-mfc'. It could take inspiration from @rgieseke's greenhouse gas renamer project

docs/parameters.rst
```diff
@@ -1,277 +1,179 @@
...
+    ``World``
+    ``World``, ``Northern Hemisphere``
+    ``World``, ``Northern Hemisphere``, ``Ocean``
+    ``World``, ``Northern Hemisphere``, ``Land``
+    ``World``, ``Southern Hemisphere``
+    ``World``, ``Southern Hemisphere``, ``Ocean``
+    ``World``, ``Southern Hemisphere``, ``Land``
+    ``World``, ``Ocean``
+    ``World``, ``Land``
+    ``World``, ``R5ASIA``
```
znicholls: can you add a TODO for me here to dig out the links of where all these come from?

docs/parameters.rst
swillner: Not in the API...

docs/parameters.rst
swillner: Determining which parametres were used should not be task of an adapter (see #123).

> If they just blow up when given a key parameter then they're useless.

Why should they?

docs/parameters.rst
swillner: Hmm, could be part of the "Description" if there are really non-trivial variants... Probably we should better properly describe HFC365mfc et al...

docs/parameters.rst
swillner: Hmm, it's the standard expression, no? "Gas pool"?

docs/parameters.rst
znicholls: We're not on the same page. 

What I want to make clear is that the user should always be able to pass key parameters e.g. `ECS` or `TCR` to a high-level `run` call and get sensible output i.e. things either run, a clear warning is raised or a clear error is raised. We don't want users passing 'standard' parameters in, only to receive a cryptic error message (which is possible if the implementer of an adapter doesn't think about how all (probably 4 or 5) of the key parameters would be handled). 

The text you've suggested does not make this clear ('as far as possible' is vague and doesn't tell the implementer what they actually have to do). The text I have clearly also fails to communicate what I wanted so I think we need to re-write this sentence somehow.

Doing this checking has to be done somewhere. I thought the most natural place was an adapter but maybe I'm missing something?

docs/parameters.rst
znicholls: definitely outdated/unthinking comment by me, oops

docs/parameters.rst
znicholls: I think so, it just seems unsatisfactory that it is so...

docs/parameters.rst
znicholls: ok I'll add in once this is merged

docs/parameters.rst
znicholls: @swillner ping

swillner: Sorry ;) Done

docs/parameters.rst
swillner: > We're not on the same page.

No ;)

> What I want to make clear is that the user should always be able to pass key parameters e.g. `ECS` or `TCR` to a high-level `run` call and get sensible output i.e. things either run, a clear warning is raised or a clear error is raised. We don't want users passing 'standard' parameters in, only to receive a cryptic error message (which is possible if the implementer of an adapter doesn't think about how all (probably 4 or 5) of the key parameters would be handled).

I don't understand why that would lead to errors at all if parameters are simply ignored by the adapter because it does not know about them. I think we must not make all adapters use every parameter described here, many models would not be able to use many of these anyway (e.g. non-carbon-gases for DICE). Adapter developers should only make sure that the parameters for the respective model are consistent with these standard parameters (e.g. giving ECS a different name).

> The text you've suggested does not make this clear ('as far as possible' is vague and doesn't tell the implementer what they actually have to do). The text I have clearly also fails to communicate what I wanted so I think we need to re-write this sentence somehow.

Maybe like the text in m,y previous reply?

> Doing this checking has to be done somewhere. I thought the most natural place was an adapter but maybe I'm missing something?

I would do that in the Core so adapters do not have to be altered when we add or change standard parameters.

znicholls: Right I get it now. Handling this in core will also work so let's go for something like

"
However, to be included in OpenSCM they should make sure their parameters fit into these standard parameters as far as possible. 
"

We can then work out the details as we start building useful tools around the core <-> adapter connection

znicholls:
@swillner hit merge when ready

PR #160, opened 2019-05-06
Add missing strip so description is right

znicholls:
Strips lines before comparison so project description is as intended

- [x] Implementation finished
- [x] Tests added (N/A)
- [x] Documentation added
- [x] Description in ``CHANGELOG.rst`` added (single line such as: ``(`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Added feature which does something``) (N/A)


setup.py
```diff
@@ -100,13 +100,15 @@
...
-        if line == ".. sec-begin-long-description":
+        if line.strip() == ".. sec-begin-long-description":
             add_line = True
-        elif line == ".. sec-end-long-description":
+        elif line.strip() == ".. sec-end-long-description":
             break
         elif add_line:
             README_LINES.append(line)
 
+assert len(README_LINES) > 2  # some description should appear
```
swillner:
```suggestion
if len(README_LINES) < 3:
    raise RuntimeError("Insufficient description given")
```

znicholls:
> Why exactly was that necessary?

Trailing whitespace meant `if line.strip() == ".. sec-begin-long-description":` wasn't being hit.

PR #163, opened 2019-05-13
Update pyam version

znicholls:
<Please replace this paragraph by a description of what this PR does and alter the TODO-list below to what applies here. Make sure the relevant points are done before requesting a merge. Please use the "wip" label if this PR is "Work in progress">

- [x] Implementation finished (N/A)
- [x] Tests added (N/A)
- [x] Documentation added (N/A)


tests/unit/test_scmdataframe/test_pyam_compat.py
```diff
@@ -11,10 +11,9 @@ def test_to_int_value_error(test_iam_df):
     bad_val = "20311/123/1"
     idf.loc[4, "time"] = bad_val
 
-    bad_vals = idf[idf["time"] == bad_val]["time"]
     error_msg = re.escape(
         "All time values must be convertible to datetime. The following values are "
-        "not:\n{}".format(bad_vals)
+        "not:\n"
```
swillner: Remove the last sentence?

CHANGELOG.rst
```diff
@@ -4,6 +4,7 @@ Changelog
 master
 ******
 
+- (`#163 <https://github.com/openclimatedata/openscm/pull/163>`_) Lock ``pyam`` version to pypi versions
```
swillner: Shouldn't we add a policy to only add bigger changes in the changelog?

znicholls: we can do, I've always just put everything in there but we can do whatever

tests/unit/test_scmdataframe/test_pyam_compat.py
znicholls: I've just removed it from the test because its unreasonably sensitivie if you test that the formatting is always identical. The error message still contains the bad values.

PR #165, opened 2019-05-22
Scenarios module

znicholls:
Adds scenario module to OpenSCM. This gives us a bunch of scenarios with which we can test models and also makes it easy for first-time users to get a feel for how OpenSCM works without having to wrangle scenario data to get started.

Also includes a minor bugfix in `ScmDataframe`'s interpolation method.

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added


znicholls:
thanks @swillner! I think I've addressed all the comments now so ready for round 2

znicholls:
we might want to re-write the timeseries notebook to make it more obvious that OpenSCM is super to easy to use, putting the details later in the notebook for those who are interested.

swillner:
I would very much like the timeseries conversion notebook stick to the internal procedure and thus not use normally exposed interfaces... Probably we should have a thorough revision of all the notebooks once the package itself is in a releasable state.

So without those changes I think we are good to go. Maybe rename the column to "ParameterType" to be consistent with the (apparent) camel case names of the other columns (or use snake_case throughout)?

znicholls:
Just to make sure I've understood, you want a notebook with only private
interfaces exposed?

Yes should be snake_case throughout, that's lazy by me.

swillner:
Yes, it's about the internals, right? Anyway, let's leave the notebooks as they are right now and properly clean them once have a release candidate.

can you then also make the other columns snake case, please? ;)

znicholls:
> Yes, it's about the internals, right?

I thought it was about the fact that there are two different conventions with respect to timeseries and that they have to be handled differently. 

> Anyway, let's leave the notebooks as they are right now and properly clean them once have a release candidate.

Ok I'll just add a note in the top saying they're likely to be re-written once we have a release and put some suggestions about how re could re-write

swillner:
Cool, just give me a ping and I'll merge ;)

znicholls:
@swillner ping. A lot of the notebook changes are unavoidable cause of the case changes and the fact we can use scenarios now rather than downloading from iiasa in examples

PR #168, opened 2019-05-24
Fix false positive detection of duplicate timeseries when appending continuous timeseries

znicholls:
Fix false positive of duplicate time points when appending timeseries

- [x] Implementation finished
- [x] Tests added
- [x] Documentation added (N/A)


znicholls:
@lewisjared one of us should fix this when we have a minute...

openscm/scmdataframe/base.py
```diff
@@ -1543,3 +1532,33 @@ def df_append(
         return ret
 
     return None
+
+
+def _handle_potential_duplicates_in_append(data, duplicate_msg):
+    # double check we're not looking at continuous timeseries
+    continuous = True
+    for _, tsdf in data.groupby(data.index.names):
```
znicholls: @lewisjared if you can see a way to avoid the for loop that would be awesome

PR #171, opened 2019-06-03
Core implementation zn

znicholls:
My updates

- [x] remove comments in dice adapter
- [x] remove `start_time` and `stop_time`
- [x] tidy up to/from parameterset conversions
    - [x] test circular conversions with generic, scalar and timeseries types
    - [x] autofill parameter_type on conversions
    - [x] make sure point/average handling is consistent/sane
- [x] make issue about handling non-World scalar/generic parameters (see #172). I think it makes sense to leave this for later because a) a solution is not obvious and b) we don't have any need for it yet


openscm/adapters/dice.py
```diff
@@ -24,49 +24,50 @@
...
     "mat0": (830.4, "GtC"),  # 851
-    # Equilibrium pool size atmosphere
+    # Equilibrium size of atmospheric CO2 pool
     #     Original: "Equilibrium concentration atmosphere (GtC)"
     "mateq": (588, "GtC"),  # 588
+    # do we know what this is?
     "mat_lower": (10, "GtC"),  # 10
     # Original: "Initial Concentration in lower strata 2010 (GtC)"
-    "ml0": (10010, "GtC"),  # 1740
+    "ml0": (10010, "GtC"),  # 1740 # do we know what this actually is?
```
swillner: let's not include those comments here

CHANGELOG.rst
```diff
@@ -25,6 +25,5 @@ master
 - (`#83 <https://github.com/openclimatedata/openscm/pull/83>`_) Rename OpenSCMDataFrame to ScmDataFrame
 - (`#78 <https://github.com/openclimatedata/openscm/pull/78>`_) Added OpenSCMDataFrame
 - (`#44 <https://github.com/openclimatedata/openscm/pull/44>`_) Add timeframes module
-- (`#41 <https://github.com/openclimatedata/openscm/pull/41>`_) Add DICE adapter
```
swillner: why remove that?

znicholls: my bad didn't realise we're doing two in one

swillner:
Nice! I like the "delta_*" unit!
Do we really want defaults for start time and stop time? I would leave those for the adapters to set

znicholls:
> Do we really want defaults for start time and stop time?

Ok so should we just ensure they're in the ParameterSet but not set their values? I wasn't really sure how we wanted to handle these standard parameters so this was just a first implementation

swillner:
And let's call them "Start Time" and "Stop Time" to be consistent.

swillner:
> Ok so should we just ensure they're in the ParameterSet but not set their values? I wasn't really sure how we wanted to handle these standard parameters so this was just a first implementation

I think raising an error when they are not set (which hapens anyway when the adapter requests them) should suffice.

znicholls:
> I think raising an error when they are not set (which hapens anyway when the adapter requests them) should suffice.

Ok. Just thinking about further, how would this work with 'equilibrium climate sensitivity'? We want to force adapters to understand and use this parameter, how do we do so?

> And let's call them "Start Time" and "Stop Time" to be consistent.

"StartTime" and "StopTime"? Consistent with what?

swillner:
> Ok. Just thinking about further, how would this work with 'equilibrium climate sensitivity'? We want to force adapters to understand and use this parameter, how do we do so?

No need to force anything: If an adapter understands ecs it needs to get it from the parameterset. If it's not there but required, an error is raised (using `view.ensure()`). As our parameter names are a convention there is no way to enforce that with tests or similar (an adapter could always use "Whatever|ECS" for ecs, but would break convention then.

> "StartTime" and "StopTime"? Consistent with what?

With the naming style of the other parameters (e.g. "Surface Temperature): "Start Time"/"Stop Time" (already have them in the docs in #146, though still with wrong units)

znicholls:
> No need to force anything

Fuck me I don't know why I can't get my head around this. Every time I misunderstand and every time it's obvious when you say it. My bad. 

The one test I think we add is that when the user puts e.g. "ECS" in their ParameterSet, the adapter knows what it should do. i.e. in the `AdapterTester` base class I'd add an abstract method, "openscm_parameters_handling" which runs the model with e.g. "Start Time", "Stop Time", "ECS" so that the implementer is forced to check that something sensible happens when these parameters are passed (it doesn't matter whether they're ignored, used, an error is raised etc. but the key is that the implementer thinks about what should happen). 

> With the naming style of the other parameters (e.g. "Surface Temperature): "Start Time"/"Stop Time" (already have them in the docs in #146, though still with wrong units)

Back on board, thought you meant variable names for some reason...

znicholls:
> With the naming style of the other parameters (e.g. "Surface Temperature): "Start Time"/"Stop Time" (already have them in the docs in #146, though still with wrong units)

I need to fix the initialisation logic to get this to work, will do tomorrow.

znicholls:
Ok have got initialisation logic 'fixed' ('fixed' as it works but I'm not sure it's the best implementation).

Also put in test of standard parameter handling for DICE.

Back to you

PR #181, opened 2019-06-11
Core implementation zn

znicholls:
@swillner this gets the test coverage as high as I can get it. A few remaining questions and then I think let's merge #146 and ignore the other PRs for now given how many issues are floating about.

### DIFF
--- a/openscm/adapters/__init__.py
+++ b/openscm/adapters/__init__.py
@@ -68,6 +68,7 @@ class Adapter(metaclass=ABCMeta):
         Called before the adapter is used in any way and at most once before a call to
         :func:`run` or :func:`step`.
         """
+        # I don't know why this isn't picked up by code coverage...
         if not self._initialized:
             self._initialize_model()
             self._initialized = True
--- a/openscm/adapters/__init__.py
+++ b/openscm/adapters/__init__.py
@@ -68,7 +68,6 @@ class Adapter(metaclass=ABCMeta):
         Called before the adapter is used in any way and at most once before a call to
         :func:`run` or :func:`step`.
         """
-        # I don't know why this isn't picked up by code coverage...
         if not self._initialized:
             self._initialize_model()
             self._initialized = True
--- a/openscm/core/__init__.py
+++ b/openscm/core/__init__.py
@@ -80,7 +80,8 @@ class OpenSCM:
         """
         Set of output parameters of the model
         """
-        return self._output_parameters
+        raise NotImplementedError  # noqa
+        # return self._output_parameters
 
     @property
     def parameters(self) -> ParameterSet:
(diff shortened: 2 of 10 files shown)


openscm/adapters/__init__.py
```diff
@@ -68,6 +68,7 @@ def initialize_model_input(self) -> None:
         Called before the adapter is used in any way and at most once before a call to
         :func:`run` or :func:`step`.
         """
+        # I don't know why this isn't picked up by code coverage...
```
swillner: Do we really want to keep these comments? ;)

openscm/core/__init__.py
```diff
@@ -80,7 +80,8 @@ def output(self) -> ParameterSet:
         """
         Set of output parameters of the model
         """
-        return self._output_parameters
+        raise NotImplementedError  # noqa
```
swillner: Why mark that as not implemented? It's just a getter...

swillner: same below...

openscm/core/parameters.py
```diff
@@ -69,7 +69,9 @@ def timeseries_type_to_string(cls, timeseries_type: "ParameterType") -> str:
             return "average"
         if timeseries_type == ParameterType.POINT_TIMESERIES:
             return "point"
-        raise ValueError("Timeseries type expected")
+        raise ValueError(
+            "Timeseries type expected"
+        )  # pragma: no cover # emergency valve
```
swillner: We can easily test this. Can you undo "no cover", please? (also below)

openscm/core/time.py
```diff
@@ -140,8 +140,8 @@ def _format_datetime(dts: np.ndarray) -> np.ndarray:
         return np.array([], dtype="datetime64[s]")
 
     dtype = dts.dtype.type
-    if issubclass(dtype, np.object):
-        dtype = np.dtype(type(dts[0])).type
+    # if issubclass(dtype, np.object):
```
swillner: Let's please leave it like before. Getting only the first element type should only be a fallback option.

tests/unit/test_core.py
```diff
@@ -169,6 +170,78 @@ def test_parameterset_named_initialization():
...
+    elif view_type == "timeseries":
+        unit = "kg"
+        tp = create_time_points(
+            np.datetime64("2010-01-01"), np.timedelta64(365, "D"), 10, "average"
+        )
+        v = paraset.timeseries(para_name, unit, tp)
+
+    str_rep = str(v)
+
+    # TODO: decide whether to unify these
```
swillner: Oh, good point. The "read-only" is a relic...

openscm/adapters/__init__.py
znicholls: Nope just making sure you've seen it

openscm/core/__init__.py
znicholls: because it's not being tested, again just to make sure that you've seen the line

tests/unit/test_core.py
znicholls: Ok I leave that for you

PR #186, opened 2019-07-02
Refactor adapters

znicholls:
<Please replace this paragraph by a description of what this PR does and alter the TODO-list below to what applies here. Make sure the relevant points are done before requesting a merge. Please use the "wip" label if this PR is "Work in progress">

- [ ] Implementation finished
- [ ] Tests added
- [ ] Documentation added
- [ ] Example added (in the documentation, to an existing notebook, or in a new notebook)
- [ ] Description in ``CHANGELOG.rst`` added (single line such as: ``(`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Added feature which does something``)


openscm/adapters/__init__.py
```diff
@@ -29,72 +31,119 @@ class Adapter(metaclass=ABCMeta):
...
 
     _output: ParameterSet
     """Output parameter set"""
 
-    _parameters: ParameterSet
-    """Input parameter set"""
+    _parameter_views: Dict[HierarchicalName, ParameterInfo]
+    """Parameter views available with this model"""
+
+    _parameter_versions: Dict[HierarchicalName, int]
```
swillner: No need for separeate versions as these are already stored in the views

openscm/adapters/__init__.py
```diff
@@ -29,72 +31,119 @@ class Adapter(metaclass=ABCMeta):
...
         """
         self._parameters = input_parameters
         self._output = output_parameters
-        self._initialized = False
-        self._initialized_inputs = False
-        self._current_time = 0
+        self._parameter_views = {}
+        self._parameter_versions = {}
+        self._direct_set = False
+        self._initialize_model()
```
swillner: Please revert back to the former model with the separate initialize methods. That is to

- separate initialization for the model itself as well as for a single run
- and for those ****** models that use file I/O so the `__init__` function does not get hold up by those file initializations

openscm/adapters/__init__.py
```diff
@@ -115,64 +164,162 @@ def step(self) -> np.datetime64:
...
+            if isinstance(v[1], dict)
+            or v[1].parameter_type not in (ParameterType.GENERIC, ParameterType.SCALAR)
+        ]
+        return generic_views + scalar_views + other_views
+
+    def _update_model_parameter(self, name: HierarchicalName) -> None:
+        para = self._parameter_views[name]
+        self._update_model(name, para)
+
+    @staticmethod  # should we unify getters?
```
swillner: Why have a separate function for that anyway? (same for setters)

openscm/adapters/__init__.py
znicholls: I think you still need them here as you need to know the version last time the adapter looked, not last time anyone looked through this view (which is also shared by the user who is setting things).

openscm/adapters/__init__.py
znicholls: To try and explain why I made the change more clearly: if you keep track of the views' versions then you can do everything in `reset`. The reason is that you only update things when you have to which means you can delay all operations as late as possible. If we wanted, we could shift initialise model into `reset` with some flag to make sure it's only done once. 

> * for those ****** models that use file I/O so the `__init__` function does not get hold up by those file initializations

Is this desirable? As the user I feel like I would want my model to be spun up (whatever that means) so that when I interact with it thereafter it's a) fast (cause it's not doing as many io steps) and b) definitely up to date with my adapter. I'm not super fussed either way but this was my thinking when I put `_initialise_model` in `__init__`.

openscm/adapters/__init__.py
znicholls: Because timeseries views require you to look at `.values` whereas scalar views require you to look at `.value` and having the check of which attribute you need to look at is a pain so I just put it in one place. If we unified the getters (i.e. made them both `.value` or `.values` we could avoid this).

PR #187, opened 2019-07-02
Timeseries view handling

### LINKED ISSUE #178: Timeseries view handling
# Problem 1

If you request a timeseries view, but don't actually set any data, you nonetheless prevent any data from being set if the time points don't line up properly. I'm not sure if this is the behaviour we want, I thought it would make more sense to only consider if the timeseries conversion will work when you realise the data.

## Current behaviour

```python
>>> import numpy as np
>>> 
>>> import openscm
>>> print(openscm.__version__)
0.1.0-alpha+143.g1921c09
>>> 
>>> from openscm.core.parameterset import ParameterSet
>>> from openscm.core.time import create_time_points
>>> 
>>> 
>>> p = ParameterSet()
>>> 
>>> tp1 = create_time_points(
...     np.datetime64("2000-01-01"),
...     np.timedelta64(365, "D"),
...     100,
...     "point",
... )
>>> v1 = p.timeseries(
...     'example',
...     's',
...     tp1,
... )
>>> 
>>> tp2 = create_time_points(
...     np.datetime64("1900-01-01"),
...     np.timedelta64(365, "D"),
...     150,
...     "point",
... )
>>> v2 = p.timeseries(
...     'example',
...     's',
...     tp2,
... )
openscm.errors.InsufficientDataError
```

## Proposed behaviour

```python
>>> p = ParameterSet()
>>> 
>>> tp1 = create_time_points(
...     np.datetime64("2000-01-01"),
...     np.timedelta64(365, "D"),
...     100,
...     "point",
... )
>>> v1 = p.timeseries(
...     'example',
...     's',
...     tp1,
... )
>>> 
>>> tp2 = create_time_points(
...     np.datetime64("1900-01-01"),
...     np.timedelta64(365, "D"),
...     150,
...     "point",
... )
>>> v2 = p.timeseries(
...     'example',
...     's',
...     tp2,
... )
>>> # no error raised, user can set data and error
>>> # would only be raised if data was insufficient to 
>>> # fill the view
```

*Note* The proposed behaviour would be in contradiction with the way we handle units, where an error is thrown immediately if you request a view with units that won't convert. I think this makes sense. If the units don't match, they'll never match. However, if you request a certain view of a timeseries then even though the data might not be there when you get the view, it might be there once you actually look at it. 

## Unit handling behaviour (proposed to remain unchanged)

```python
>>> p = ParameterSet()
>>> 
>>> v1 = p.scalar(
...     'example_scalar',
...     's',
... )
>>> 
>>> v2 = p.scalar(
...     'example_scalar',
...     'm',
... )
pint.errors.DimensionalityError: Cannot convert from 'second' ([time]) to 'meter' ([length])
```

# Problem 2

When you request a timeseries view, you have to specify time points. Should we have a way to simply register a timeseries' units and type, without specifying the time points. This would ensure that any future writing/view requests matched the required units/type.

## Proposed behaviour

```python
>>> p = ParameterSet()
>>> 
>>> v1 = p.timeseries(
...     'example',
...     's',
...     timeseries_type="point"
... )
>>> # view is empty to request to values raises error
>>> v1.values
openscm.errors.ParameterEmptyError
>>> # now requesting a view with unit conflict causes error
>>> v3 = p.timeseries(
...     'example',
...     'kg',
...     timeseries_type="point"
... )
pint.errors.DimensionalityError: Cannot convert from 'second' ([time]) to 'kg' ([mass])
>>> # now requesting a view with parameter type conflict causes error
>>> v4 = p.timeseries(
...     'example',
...     's',
...     timeseries_type="average"
... )
openscm.errors.ParameterTypeError  # at least until we get the point <-> average conversion working
>>> # requesting a compatible view would then let the user set the data
>>> tph = create_time_points(
...     np.datetime64("2000-01-01"),
...     np.timedelta64(365, "D"),
...     3,
...     "point",
... )
>>> v2 = p.timeseries(
...     'example',
...     'day',
...     tph,
...     timeseries_type="point",
... )
>>> v2.values = np.array([1, 2, 3])
```

znicholls:
Updates timeseries view handling to match #178 and fixes bug in time limits overlap checking

- [x] Check with @swillner about how caching is meant to behave
- [ ] Implementation finished
- [x] Tests added
- [x] Documentation added

Closes #178


znicholls:
@swillner can you review please? I don't understand how the caching is meant to work so have just broken it, your thoughts on how to fix it would be great!

openscm/core/parameterset.py
```diff
@@ -215,17 +216,37 @@ def timeseries(  # pylint: disable=too-many-arguments
...
-        return TimeseriesView(
-            parameter, unit, time_points, timeseries_type, interpolation, extrapolation
-        )  # TimeseriesView
+        if parameter.unit is not None:
+            # check conversion can be done
+            UnitConverter(parameter.unit, unit)
+
+        if time_points is None:
+            # fill with junk as can be overwritten later anyway
+            _time_points = np.array(
```
swillner: Just leave it `None`, that's also checked below

openscm/core/parameterset.py
```diff
@@ -215,17 +216,37 @@ def timeseries(  # pylint: disable=too-many-arguments
...
             name, self._get_or_create_region(region)
         )
 
-        parameter.attempt_read(timeseries_type, unit, time_points)
-        return TimeseriesView(
-            parameter, unit, time_points, timeseries_type, interpolation, extrapolation
-        )  # TimeseriesView
+        if parameter.unit is not None:
+            # check conversion can be done
+            UnitConverter(parameter.unit, unit)
```
swillner: Should not be necessary, is done in `TimeseriesView` (which should defenitely be returned in any case)

openscm/core/parameterset.py
```diff
@@ -215,17 +216,37 @@ def timeseries(  # pylint: disable=too-many-arguments
...
+            return TimeseriesView(
+                parameter,
+                unit,
+                time_points,
+                timeseries_type,
+                interpolation,
+                extrapolation,
+            )  # TimeseriesView
+
+        return ParameterInfo(parameter)  # no time points so can't get values
```
swillner: This should actually *always* return a `TimeseriesView`. That should then handle potentially `None` `time_points`.

openscm/core/time.py
```diff
@@ -422,8 +422,11 @@ def __init__(  # pylint: disable=too-many-arguments
         self._interpolation_type = interpolation_type
         self._extrapolation_type = extrapolation_type
 
-        if self._source[0] > self._target[1]:  # TODO: consider extrapolation type
-            raise InsufficientDataError
+        if extrapolation_type == ExtrapolationType.NONE:
+            if self._source[0] > self._target[-1]:
```
swillner: for ` ExtrapolationType.NONE` not a single missing time step can be extrapolated, so the check should stay as `self._source[0] > self._target[1]`, the other should be for `self._source[-1] < self._target[-2]`

openscm/core/views.py
```diff
@@ -303,64 +315,46 @@ def __init__(  # pylint: disable=too-many-arguments
...
-
-        if self._parameter.children:
-            self._child_data_views = get_data_views_for_children_or_parameter(
-                self._parameter
-            )
+        self.time_points = time_points
 
     def _read(self):
-        if self._data is None:
+        if self._data is None or self._version != self._parameter.version:
```
swillner: A new `TimeseriesConverter` should only be created if the `time_points` in the parameter changed

swillner: Doing to on every version change is a big performance hit and goes contrary to the idea of the converters and views.

openscm/core/views.py
```diff
@@ -303,64 +315,46 @@ def __init__(  # pylint: disable=too-many-arguments
...
-            )
-            self._writable = True
+        self._parameter.attempt_write(
+            self._timeseries_type, self._unit_converter.target, self._time_points
+        )
+        self._writable = True
 
     def _write(self):
         if not self._locked:
+            self._timeseries_converter = TimeseriesConverter(
```
swillner: Same here, only if `time_points` change

openscm/core/views.py
```diff
@@ -402,16 +396,24 @@ def values(self) -> _Timeseries:
             Lengths of set value and the time points number mismatch
         """
         self._read()
-        if self._timeseries is None:
-            self._timeseries = _Timeseries(self._data, self)
-        return self._timeseries
+        # if self._timeseries is None:  # @swillner I've broken the logic here somehow
```
swillner: Uff, I think the whole system is broken here. Is having overwritable time points urgent actually?

openscm/core/views.py
```diff
@@ -421,6 +423,48 @@ def values(self, v: np.ndarray) -> None:
...
+
+    @time_points.setter
+    def time_points(self, v: np.ndarray) -> None:
+        """
+        Time points of this view
+        """
+        self._time_points = v
+        self._version -= 1
+
+        def get_data_views_for_children_or_parameter(
```
swillner: Why move it here?

openscm/core/views.py
```diff
@@ -433,7 +477,7 @@ def length(self) -> int:
         """
         Length of timeseries
         """
-        return self._timeseries_converter.target_length
+        return self._timeseries_converter.target_length  # type: ignore
```
swillner: Why these `type: ignore`s? What has changed that they are necessary? Maybe use a `cast` instead? Same below...

openscm/core/parameterset.py
znicholls: Ok. I'll need to modify `TimeseriesConverter` so it can take time points of none then (or think of some other solution).

openscm/core/parameterset.py
znicholls: As above, I'll need to modify TimeseriesConverter so it can take time points of none or modify `TimseriesView` so it just doesn't have a `TimeseriesConverter` if `time_points` is `None`.

openscm/core/parameterset.py
znicholls: Ok cool, think that solves above two comments too then.

openscm/core/time.py
znicholls: Should it not be self._source[0] > self._target[0] then? (I'm missing something but I can't see what...)

openscm/core/views.py
znicholls: Yep nice

openscm/core/views.py
znicholls: Nice

openscm/core/views.py
znicholls: It makes life a lot easier in #186 so I'd like to spend a bit more time on it. I think we can do this without destroying everything, I just needed your help to see how (and think I can work it out now...)

openscm/core/views.py
znicholls: It was the change to returning `ParameterInfo` instead of `TimeseriesView`. When I get rid of that, these will disappear.

openscm/core/views.py
znicholls: Because the child_views depend on the `time_points` and so have to all be updated whenever the `time_points` are updated.

znicholls:
Superceded by #189

PR #198, opened 2019-09-23
jupyter notebook for comparison of fair and magicc

rbeddig:
Jupyter notebook for the comparison of the simple climate models FaIR and MAGICC using packages fair and pymagicc.

- [ ] Implementation finished
- [ ] Tests added
- [ ] Documentation added
- [ ] Example added (in the documentation, to an existing notebook, or in a new notebook)
- [ ] Description in ``CHANGELOG.rst`` added (single line such as: ``(`#XX <https://github.com/openclimatedata/openscm/pull/XX>`_) Added feature which does something``)


znicholls:
Cheers @rbeddig! It will probably be a little while before we can have a look through this but will be helpful documentation when we get there.

Issue #183, opened 2019-06-27
ScmDataFrame: Default to `np.datetime64` for internal date handling


swillner:
Currently `ScmDataFrame` uses `datetime.datetime` to represent dates internally. As the rest of OpenSCM uses `np.datetime64`, this could/should be replaced by `np.datetime64`.

OlgaCh:
Hi @swillner 
May I grab this one as a part of `Hacktoberfest` contribution?

OhYash:
hey @OlgaCh, are you working on this?

OlgaCh:
@OhYash no, I don't.

swillner:
Hey @OlgaCh , very sorry I haven't replied yet. Please, go ahead, we would be very happy!

AlbertUlysses:
hello @swillner @OlgaCh curious to know if this has been done, otherwise I would love to help!

OhYash:
@AlbertUlysses Doesn't look to be done, have you started with this?

AlbertUlysses:
I started looking at the code, yes. I think I know where the change should be. @OhYash

dashuber:
Looks a lot like this issue's still open. @swillner see #199

PR #199, opened 2019-10-27
Change datetime type to np.datetime64

dashuber:
I changed the datetime.datetime typ to np.datetime64 in the scmdataframe module. Did I forget something? Please let me know if the PR isn't to your liking.

### DIFF
--- a/openscm/scmdataframe/base.py
+++ b/openscm/scmdataframe/base.py
@@ -184,7 +184,7 @@ def _format_wide_data(df):
     time_cols, extra_cols = False, []
     for i in cols:
         # if in wide format, check if columns are years (int) or datetime
-        if isinstance(i, datetime.datetime):
+        if isinstance(i, np.datetime64):
             time_cols = True
         else:
             try:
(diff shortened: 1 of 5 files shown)


openscm/scmdataframe/base.py
```diff
@@ -194,7 +194,7 @@ def _format_wide_data(df):
                 try:
                     try:
                         # most common format
-                        datetime.datetime.strptime(i, "%Y-%m-%d %H:%M:%S")
+                        pd.to_datetime(i, format="%Y-%m-%d %H:%M:%S")
```
znicholls: hi @dashuber, we currently avoid using `pd.to_datetime` because of its time bounds limitations (see https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#timestamp-limitations). It means you get errors like the below for time ranges we want to use

```python
>>> import pandas as pd
>>> pd.to_datetime("2500-02-10 12:00:00")
pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 2500-02-10 12:00:00
```

znicholls:
@lewisjared this closes #183

PR #203, opened 2020-01-15
Fix pandas version requirement

zikolach:
Minor change as originally referenced issue IAMconsortium/pyam#245 was closed and latest pyam requires pandas >= 25.0 (despite of declared pandas>0.24)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ REQUIREMENTS_INSTALL = [
     "scmdata",
     "scipy",
     "pint",
-    "pandas>=0.24.2",
+    "pandas>=0.25",
     "python-dateutil",
     "tqdm",
 ]
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ REQUIREMENTS_NOTEBOOKS = [
     "matplotlib",
     "notebook",
     "seaborn",
-    "pyam-iamc>=0.2.0",
+    "pyam-iamc>=0.4.0",
     "ipywidgets",
 ]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -41,7 +41,7 @@
     "scmdata",
     "scipy",
     "pint",
-    "pandas==0.24.2",  # Until next release https://github.com/IAMconsortium/pyam/issues/245
+    "pandas>=0.24.2",
```
znicholls:
```suggestion
    "pandas>=0.25",
```

can you also pin the pyam version too please?

znicholls:
thanks @zikolach ! One tiny change then should be good to go?

setup.py
zikolach: I just set it to `>=0.4.0`

znicholls: Pandas needs to be pinned too right? Or is that pinning done by pyam?

znicholls: If it’s done by pyam we can delete the pandas pinning then right?

znicholls:
thanks @zikolach !!

setup.py
zikolach: The thing is in just released pyam (0.4.0) there is wrong pandas requirement (it requires >= 0.25 to pass all tests). I [updated](https://github.com/IAMconsortium/pyam/commit/957af4b08a567ebf9e960d3822db160af1a4783e) it yesterday but it was after release 0.4.0.

znicholls: Ah ok cool perfect then

